PR #13, opened 2020-12-04
Implement example of multi-item dragging

henryyi:


henryyi:
Had rebased and force pushed since there needed to be many changes to get the PR up to date.

MasNotsram:
@slugmandrew Seconded!

clauderic:
This isn't merged because:

- There's a newer PR in favour of which this PR will be closed that implements the same thing here: https://github.com/clauderic/dnd-kit/pull/588
- It significantly increases the overall complexity of the stories, which are already very complicated to understand for new consumers. I'm not planning on merging https://github.com/clauderic/dnd-kit/pull/588 either until the complexity of the stories is addressed.

slugmandrew:
Fair enough... does seem a shame to miss out on this kind of community contribution though, especially as people are clearly putting a lot of time into trying to make the library better and it drives engagement.

Maybe I'm missing some discourse elsewhere (?), but doesn't even seem like you said this here and it was just left hanging.

#588 Looks like a really nice solution though, and I agree the code is cleaner.

MuhammadHamzaChippa:
Did anyone try selecting items from different containers?

script8888:
How can I use this multi select pls?
I'm lost

can anyone please help out?

ricardomejiasilva:
Here is an implementation I came up with using multiple containers 
https://github.com/ricardomejiasilva/drag-drop-spec-tables

AndrejGajdos:
@ricardomejiasilva thanks, I see you made multidragging up and running

PR #37, opened 2021-01-05
Fix URL to GitHub repository

nickpresta:
Hi 👋 

I was viewing this package on [BundlePhobia](https://bundlephobia.com/result?p=@dnd-kit/core@1.0.0) and tried to jump to GitHub directly but noticed it was linking to `https://github.com/claudric/dnd-kit`

This change fixes the GitHub URL, which will get picked up in the next version bump.

Have a great day 🙂


clauderic (approved):
:yells-at-broken-macbook-keyboard:

PR #54, opened 2021-01-13
[WIP] Update scroll ancestor detection when the draggable is not over a droppable

### LINKED ISSUE #43: RectIntersection is not working with draggable items inside a scrollable container
The draggable items that are scrolled are not calculating the `intersectionRatio` correctly.

Please, see this code sandbox for an example: https://codesandbox.io/s/react-dnd-grid-0ylzl?file=/src/App.js

NOTE: Try dragging the last draggable item into one droppable area.

clauderic:
This PR is currently a work in progress, but will eventually resolve #43.

Currently, if the active draggable is not over a droppable, the scroll ancestor logic assumes that the scroll ancestors should be those of the active draggable.

This behaviour is problematic for a number of reasons, including those outlined in #43.

An alternative approach that I'm exploring in this PR is to find the scroll element automatically based on the center point of the active draggable. This isn't a perfect solution, which is why I'm opening this PR as a draft PR for the time being.

This solution isn't perfect currently because the point coordinates used for scroll ancestor detection should be affected by modifiers, but the modifiers also need the scroll ancestors. One solution for this that I'm currently thinking of would be to distinguish between the scroll target ancestors and the active node scroll ancestors.

Also, the `document.getElementFromPoint` method is expensive, and therefore calls to it should be debounced to avoid performance issues. Finding a scroll ancestor when the draggable is not over a droppable is something that can be done asynchronously and doesn't necessarily need to happen immediately for every drag move event.

### DIFF
--- a/packages/core/src/components/DndContext/DndContext.tsx
+++ b/packages/core/src/components/DndContext/DndContext.tsx
@@ -32,6 +32,7 @@ import {
   useScrollableAncestors,
   useClientRect,
   useClientRects,
+  useFindElementFromPoint,
   useScrollOffsets,
   useViewRect,
   SyntheticListener,
@@ -50,6 +51,7 @@ import {
   CollisionDetection,
   defaultCoordinates,
   getAdjustedRect,
+  centerOfRectangle,
   getRectDelta,
   rectIntersection,
 } from '../../utilities';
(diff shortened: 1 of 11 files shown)


py-wai:
Hi @clauderic  👋

Love your library, huge work!

This PR concerns the Sortable preset as well?

clauderic:
Hey @py-wai, the sortable preset is built on top of `@dnd-kit/core` and thus benefits from any updates/fixes made to the core lib

arpithindukuri:
@clauderic thank you for the fantastic library and great support.

I have faced similar issues to #43, and I am happy to try and help solve this.

I am slowly learning the inner workings of this library, so is there any way I can help with this PR?

clauderic:
Closing in favor of https://github.com/clauderic/dnd-kit/pull/518

Issue #105, opened 2021-02-27
DragOverlay - Cannot read property 'then' of undefined


oliverfindl:
Hello,

I was trying to fiddle with this package and followed instructions for [Sortable example](https://docs.dndkit.com/presets/sortable). At [final step](https://docs.dndkit.com/presets/sortable#drag-overlay), after adding `DragOverlay` it crashes.

![image](https://user-images.githubusercontent.com/33978096/109389903-d4b3fa80-790e-11eb-8a7d-37cd0a6fc08a.png)

After removing `DragOverlay` it works again. 

Couldn't find this bug in issues, so I'm reporting it.

Thanks.

ranbena:
@oliverfindl a few questions:
1. Do the [Storybook Sortable Examples](https://5fc05e08a4a65d0021ae0bf2-zditmhymjz.chromatic.com/?path=/story/presets-sortable-vertical--basic-setup) generate the same error for you?
  If not, the reason for the error might be in the way you implemented it. Perhaps share a reproducible code example.

2. What browser version are you testing with? Some do not support [Animation.finished](https://developer.mozilla.org/en-US/docs/Web/API/Animation/finished) which dnd-kit relies on.

oliverfindl:
Hello,

Storybook example is also broken for me.

I'm using opera 68.0.3618.173 (latest version in repo of my OS, but not latest version of Opera) that should be equivalent to Chrome 81.0.4044.138. Just tested in newest Chrome and Firefox, and its working there. According [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Animation/finished#browser_compatibility) and [CIU](https://caniuse.com/mdn-api_animation_finished), `Animation.finished` is not even supported in Chrome (although its working for me), so guess, its not ready for production use (yet).

Anyway, I found out [this polyfill](https://github.com/web-animations/web-animations-js) which solved issue for me.

Closing this issue.

Thanks.

PR #106, opened 2021-02-27
Improve animation finish browser support

ranbena:
Addresses issue #105.

### The bug
Dnd-kit [relies](https://github.com/clauderic/dnd-kit/blob/master/packages/core/src/components/DragOverlay/hooks/useDropAnimation.ts#L101) on [animation.finished](https://developer.mozilla.org/en-US/docs/Web/API/Animation/finished), which [lacks in browser support](https://caniuse.com/mdn-api_animation_finished).
I encountered this error on Electron v9.4.1 (Electron supports `animation.finished` only from version 10.x).

### Proposed solution
Swap it with [animation.onfinish](https://developer.mozilla.org/en-US/docs/Web/API/Animation/onfinish), which has [better browser supported](https://caniuse.com/mdn-api_animation_onfinish).

### DIFF
--- a/.changeset/animation-finish-detection.md
+++ b/.changeset/animation-finish-detection.md
@@ -1,0 +1,5 @@
+---
+"@dnd-kit/core": patch
+---
+
+Replace `animation.finished` with `animation.onfinish` for DragOverlay drop animation as the latter has much better support across browsers. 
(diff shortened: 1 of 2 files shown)


clauderic (approved):
LGTM 👍

Issue #16, opened 2020-12-04
DragOverlay does not change position when the window is resized


husshegz:
Go inside the storybook - Multiple containers story

1) Focus on one item with your keyboard (tab+space)
2) resize window
3) Item overflows

PR #111, opened 2021-03-01
Cancel drag on window resize

ranbena:
Fixes issue #16.

Knowing this issue is an edge case, working on updating the draglayer position to match the viewport size might be unnecessarily complex. A simpler approach is to cancel the drag when the window is resized.

Here's how it looks:

https://user-images.githubusercontent.com/486954/109474789-8a9d5700-7a7d-11eb-9cd2-3a5a3ba60ef1.mov

### DIFF
--- a/packages/core/src/sensors/keyboard/KeyboardSensor.ts
+++ b/packages/core/src/sensors/keyboard/KeyboardSensor.ts
@@ -31,6 +31,7 @@ export class KeyboardSensor implements SensorInstance {
   public autoScrollEnabled = false;
   private coordinates: Coordinates = defaultCoordinates;
   private listeners: Listeners;
+  private windowListeners: Listeners;
 
   constructor(private props: KeyboardSensorProps) {
     const {
@@ -39,7 +40,9 @@ export class KeyboardSensor implements SensorInstance {
 
     this.props = props;
     this.listeners = new Listeners(getOwnerDocument(target));
+    this.windowListeners = new Listeners(window);
     this.handleKeyDown = this.handleKeyDown.bind(this);
+    this.handleCancel = this.handleCancel.bind(this);
 
     this.attach();
   }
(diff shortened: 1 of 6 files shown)


packages/core/src/sensors/keyboard/KeyboardSensor.ts
```diff
@@ -39,15 +40,20 @@ export class KeyboardSensor implements SensorInstance {
...
     this.attach();
   }
 
   private attach() {
     this.handleStart();
 
-    setTimeout(() => this.listeners.add('keydown', this.handleKeyDown));
+    setTimeout(() => {
+      this.listeners.add('keydown', this.handleKeyDown);
+      this.windowListeners.add('resize', this.handleCancel);
```
ranbena, line 55: Can instead just directly add listener to window without the `new Listeners(window)` instance.
```ts
window.addEventListener('resize', this.handleCancel);
```
@clauderic Whatever you think is best.

packages/core/src/sensors/keyboard/KeyboardSensor.ts
```diff
@@ -39,15 +40,20 @@ export class KeyboardSensor implements SensorInstance {
 
     this.props = props;
     this.listeners = new Listeners(getOwnerDocument(target));
+    this.windowListeners = new Listeners(window);
```
clauderic, line 43: This would need to be the window of the event target
```suggestion
    this.windowListeners = new Listeners(getWindow(target));
```

packages/core/src/sensors/keyboard/KeyboardSensor.ts
clauderic: I think what you have already is good, since that same pattern is used across all sensors it's best to also use it here for the sake of consistency

clauderic (approved):

PR #123, opened 2021-03-09
Added mouse move debouncing

ranbena:
While working with dnd-kit I noticed occasional janky dragging (I'm using a custom collision detector).

As a performance optimization, mouse move handler execution can be debounced with `requestAnimationFrame`, as it is redundant to do so at a higher rate than the screen's fps.

### DIFF
--- a/cypress/support/commands.ts
+++ b/cypress/support/commands.ts
@@ -68,6 +68,7 @@ Cypress.Commands.add(
               initialRect.top + initialRect.height / 2 + y / 2
             ),
           })
+          .wait(60)
           .trigger('mousemove', {
             force: true,
             clientX: Math.floor(initialRect.left + initialRect.width / 2 + x),
--- a/packages/core/src/sensors/pointer/AbstractPointerSensor.ts
+++ b/packages/core/src/sensors/pointer/AbstractPointerSensor.ts
@@ -2,7 +2,11 @@ import {subtract as getCoordinatesDelta} from '@dnd-kit/utilities';
 
 import {getEventListenerTarget, Listeners} from '../utilities';
 
-import {getEventCoordinates, getOwnerDocument} from '../../utilities';
+import {
+  getEventCoordinates,
+  getOwnerDocument,
+  rAFDebounce,
+} from '../../utilities';
 import {KeyboardCode} from '../keyboard';
 import type {SensorInstance, SensorProps, SensorOptions} from '../types';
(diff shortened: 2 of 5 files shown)


ranbena:
@clauderic lmk what you think of this PR and approach. I can make a before/after demo if necessary.

Also, I'm thinking of taking it a step further by separating `<DragOverlay>` and `onMove` in a way that each gets its own debouncer (or perhaps making `onMove` async) -  resulting in an ever-responsive drag overlay (never lagging behind the mouse cursor), regardless of a busy `onMove` handler. Wdyt?

clauderic:
I'm not completely opposed to this change, but I'm a bit hesitant to add `requestAnimationFrame` to such a core component of the library. Browsers can heavily throttle `requestAnimationFrame` requests in a number of different scenarios, which may make the drag and drop interactions unresponsive.

Is the janking caused by the complexity of your custom collision detection algorithm? Could you perhaps debounce the collision detection function with `requestAnimationFrame` instead?

ranbena:
> Browsers can heavily throttle requestAnimationFrame requests in a number of different scenarios, which may make the drag and drop interactions unresponsive.

I personally have never had issues with raf throughout the years.
My guess is that the browser would throttle it only if the main thread is at high cpu usage, which would probably affect dnd responsiveness anyhow. Nevertheless I understand the caution and am ok with closing the PR.

> Is the janking caused by the complexity of your custom collision detection algorithm? Could you perhaps debounce the collision detection function with requestAnimationFrame instead?

Dunno the origin of the jank is but definitely I can move the debouncing to the custom collision function 👍

clauderic:
> I personally have never had issues with raf throughout the years.
My guess is that the browser would throttle it only if the main thread is at high cpu usage, which would probably affect dnd responsiveness anyhow. Nevertheless I understand the caution and am ok with closing the PR.


You can refer to https://mattperry.is/writing-code/browsers-may-throttle-requestanimationframe-to-30fps to get an idea of some of the situations where the browser may throttle `requestAnimationFrame`.

My main concern here is just that a lot of interactions of the library devolve from this move handler, so I'm wondering if this is something that can scoped down to a more specific area of the core package.

ranbena:
I tried moving the debouncing to the collision detection, and it works well but it's still not ideal - I think it would be optimal to decouple the drag overlay position update from the collision detection (by making it async).

I'll close this for now till I get time to work on it.

PR #132, opened 2021-03-14
Added onDragEndConfirm to DndContext

ranbena:
In #107 a confirm dialog on drop was added to storybook.
The demo was implemented with `window.confirm` which is rarely used. Usually a modal is used, but since it can't stop javascript execution the way `window.confirm` does - it won't work, as the drag end handler and its drop animation just won't wait.

### My proposition
`<DndContext>` exposes a new prop `onDragEndConfirm`. If provided, it awaits a `Promise` execution in the drag end handler, expecting a boolean return value:
* `true` continues `onDragEnd`
* `false` aborts `onDragEnd` and triggers `onDragCancel`

```tsx
<DndContext onDragEndConfirm={(event: onDragEvent) => Promise<boolean>}>
```

### Screencast

https://user-images.githubusercontent.com/486954/111068510-b3880800-84d1-11eb-9c6c-b7866a9b4809.mov

### DIFF
--- a/packages/core/src/components/DndContext/DndContext.tsx
+++ b/packages/core/src/components/DndContext/DndContext.tsx
@@ -125,6 +125,7 @@ interface Props {
   onDragMove?(event: DragMoveEvent): void;
   onDragOver?(event: DragOverEvent): void;
   onDragEnd?(event: DragEndEvent): void;
+  onDragEndConfirm?: (event: DragEndEvent) => Promise<boolean>;
   onDragCancel?(event: DragCancelEvent): void;
 }
 
@@ -323,13 +324,31 @@ export const DndContext = memo(function DndContext({
       setActivatorEvent(event.nativeEvent);
 
       function createHandler(type: Action.DragEnd | Action.DragCancel) {
-        return function handler() {
-          const {overId, scrollAdjustedTransalte} = tracked.current;
+        return async function handler() {
           const props = latestProps.current;
+          const {overId, scrollAdjustedTransalte} = tracked.current;
           const activeId = activeRef.current;
 
           if (activeId) {
             activeRef.current = null;
+          }
+
+          if (props.onDragEndConfirm && activeId) {
+            const confirmed = await props.onDragEndConfirm({
+              active: {
+                id: activeId,
+              },
+              delta: scrollAdjustedTransalte,
+              over: overId
+                ? {
+                    id: overId,
+                  }
+                : null,
+            });
+
+            if (!confirmed) {
+              type = Action.DragCancel;
+            }
           }
 
           dispatch({type});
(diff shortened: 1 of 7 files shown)


ranbena:
@clauderic lmk what you think. I believe this implementation adds no risk and works well.
It's important to me as it is a must-have feature in our app, and imo a common enough use case.

I can add a cypress test upon your approval.

packages/core/src/components/DndContext/DndContext.tsx
```diff
@@ -332,6 +333,24 @@ export const DndContext = memo(function DndContext({
             activeRef.current = null;
           }
 
+          if (props.onDragEndConfirm && activeId) {
```
clauderic, line 336: Should this code run if `type === Action.DragCancel`?

packages/core/src/components/DndContext/DndContext.tsx
```diff
@@ -332,6 +333,24 @@ export const DndContext = memo(function DndContext({
...
+            const confirmed = await props.onDragEndConfirm({
+              active: {
+                id: activeId,
+              },
+              delta: scrollAdjustedTransalte,
+              over: overId
+                ? {
+                    id: overId,
+                  }
+                : null,
```
clauderic, line 346: Can this object be stored in a variable higher up in the method so it can be re-used below around L363-372?

clauderic (commented):
Hey @ranbena,

Thanks for your contribution ❤️ I think overall this use case is fairly common and should be supported, I'm just wondering if `onDragEndConfirm` is the right model around this.

Would something a bit more generic such as `canDrop` make more sense? Where `canDrop` can be of type:

```ts
type CanDropHandler = (args: CanDropArguments) => (boolean | Promise<boolean>)
```

packages/core/src/components/DndContext/DndContext.tsx
ranbena: Right, I'll fix that.

packages/core/src/components/DndContext/DndContext.tsx
ranbena: Np

ranbena: To accomplish it, I had to add in `if (!activeId) return` at the top, cuz a `DragEndEvent` mandates it.
Will that have any unwanted affect?

ranbena:
@clauderic I pushed the requested changes. Take a gander.
And feel free to modify the code as you see fit before merging.

Lmk if you think it deserves an e2e test (but perhaps in subsequent PR bc some changes are needed for Cypress to be able to to flip on the storybook control).

clauderic (approved):
After giving this a bit more thought, I've renamed the prop to `cancelDrop`. Since the library doesn't actually move items from one droppable to another for consumers, the name `canDrop` felt a bit misleading for consumers. Naming is hard, but I believe the name `cancelDrop` better conveys what happens when returning `true/false` in the `canDrop` handler.

PR #140, opened 2021-03-18
Autoscrolling improvements

### LINKED ISSUE #23: Improve auto-scrolling for small containers / large drag sources
Currently if you have a small scroll container and a large drag source, it's nearly impossible to get fine controlled auto-scrolling working properly given the existing auto-scrolling logic.

The [getScrollDirectionAndSpeed](https://github.com/clauderic/dnd-kit/blob/master/packages/core/src/utilities/getScrollDirectionAndSpeed.ts) will need to be updated to take into account the relative size of the item being dragged compared to the scroll container's actual size

clauderic:
Fixes #23 

Auto-scrolling defaults have been updated, which should generally lead to improved user experience for most consumers.

The auto-scroller now bases its calculations based on the position of the pointer rather than the edges of the draggable element's rect by default. This change is aligned with how the native HTML 5 Drag & Drop auto-scrolling behaves.

This behaviour can be customized using the `activator` option of the `autoScroll` prop:

```tsx
import {AutoScrollActivator, DndContext} from '@dnd-kit/core';

<DndContext autoScroll={{activator: AutoScrollActivator.DraggableRect}} />;
```

The auto-scroller now also looks at scrollable ancestors in order of appearance in the DOM tree, meaning it will first attempt to scroll the window, and narrow its focus down rather than the old behaviour of looking at scrollable ancestors in order of closeness to the draggable element in the DOM tree (reversed tree order).

This generally leads to an improved user experience, but can be customized by passing a configuration object to the `autoScroll` prop that sets the `order` option to `TraversalOrder.ReversedTreeOrder` instead of the new default value of `TraversalOrder.TreeOrder`:

```tsx
import {DndContext, TraversalOrder} from '@dnd-kit/core';

<DndContext autoScroll={{order: TraversalOrder.ReversedTreeOrder}} />;
```

The autoscrolling `thresholds`, `acceleration` and `interval` can now also be customized using the `autoScroll` prop:

```tsx
import {DndContext} from '@dnd-kit/core';

<DndContext
  autoScroll={{
    thresholds: {
      // Left and right 10% of the scroll container activate scrolling
      x: 0.1,
      // Top and bottom 25% of the scroll container activate scrolling
      y: 0.25,
    },
    // Accelerate slower than the default value (10)
    acceleration: 5,
    // Auto-scroll every 10ms instead of the default value of 5ms
    interval: 10,
  }}
/>;
```

Finally, consumers can now conditionally opt out of scrolling certain scrollable ancestors using the `canScroll` option of the `autoScroll` prop:

```tsx
import {DndContext} from '@dnd-kit/core';

<DndContext
  autoScroll={{
    canScroll(element) {
      if (element === document.scrollingElement) {
        return false;
      }

      return true;
    },
  }}
/>;
```

### DIFF
--- a/packages/core/src/hooks/utilities/useAutoScroller.ts
+++ b/packages/core/src/hooks/utilities/useAutoScroller.ts
@@ -68,18 +68,22 @@ export function useAutoScroller({
         pointerCoordinates
       );
 
-      scrollSpeed.current = speed;
-      scrollDirection.current = direction;
-
-      clearAutoScrollInterval();
-
       if (speed.x > 0 || speed.y > 0) {
+        clearAutoScrollInterval();
+
         scrollContainerRef.current = scrollContainer;
         setAutoScrollInterval(autoScroll, interval);
 
-        break;
+        scrollSpeed.current = speed;
+        scrollDirection.current = direction;
+
+        return;
       }
     }
+
+    scrollSpeed.current = {x: 0, y: 0};
+    scrollDirection.current = {x: 0, y: 0};
+    clearAutoScrollInterval();
   }, [
     autoScroll,
     clearAutoScrollInterval,
--- a/packages/core/src/hooks/utilities/useAutoScroller.ts
+++ b/packages/core/src/hooks/utilities/useAutoScroller.ts
@@ -1,4 +1,4 @@
-import {useCallback, useEffect, useRef} from 'react';
+import {useCallback, useEffect, useMemo, useRef} from 'react';
 import {useInterval} from '@dnd-kit/utilities';
 
 import {getScrollDirectionAndSpeed, defaultCoordinates} from '../../utilities';
@@ -43,6 +43,10 @@ export function useAutoScroller({
 
     scrollContainer.scrollBy(scrollLeft, scrollTop);
   }, []);
+  const reversedScrollAncestors = useMemo(
+    () => [...scrollableAncestors].reverse(),
+    [scrollableAncestors]
+  );
 
   useEffect(() => {
     if (disabled || !scrollableAncestors.length || !pointerCoordinates) {
@@ -50,7 +54,7 @@ export function useAutoScroller({
       return;
     }
 
-    for (const scrollContainer of scrollableAncestors) {
+    for (const scrollContainer of reversedScrollAncestors) {
       const index = scrollableAncestors.indexOf(scrollContainer);
       const scrolllContainerRect = scrollableAncestorRects[index];
 
@@ -84,6 +88,7 @@ export function useAutoScroller({
     pointerCoordinates,
     setAutoScrollInterval,
     scrollableAncestors,
+    reversedScrollAncestors,
     scrollableAncestorRects,
   ]);
 }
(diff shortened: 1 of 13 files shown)


ranbena:
@clauderic I want to try this branch out in my project that uses dnd-kit, but not sure how to point to it, instead of the installed npm `@dnd-kit/core` package.
How does one achieve this? With `yarn link`? With a lerna command?

clauderic:
You can try using `yalc` or just straight up building the branch locally and replacing the distributed assets in your project's local `node_modules` if you just want to quickly test without setting up anything complicated.

This PR is still a bit of a work in progress, the acceleration and thresholds need a bit of polish. Also considering letting consumers pass in their own autoscroller offsets

packages/core/src/hooks/utilities/useAutoScroller.ts
```diff
@@ -43,14 +47,18 @@ export function useAutoScroller({
...
     scrollContainer.scrollBy(scrollLeft, scrollTop);
   }, []);
+  const sortedScrollableAncestors = useMemo(
+    () => sortScrollableAncestors(scrollableAncestors),
+    [sortScrollableAncestors, scrollableAncestors]
+  );
 
   useEffect(() => {
-    if (disabled || !scrollableAncestors.length || !draggingRect) {
+    if (disabled || !scrollableAncestors.length || !pointerCoordinates) {
```
mitchheddles, line 56: I don't know if it's intended, but it looks possible that `sortScrollableAncestors` could remove elements. Does it make sense the check the sorted ancestors list instead?

```suggestion
    if (disabled || !sortedScrollableAncestors.length || !pointerCoordinates) {
```

clauderic: Yeah, I realized that could be an unintended consequence of giving consumers a method as flexible as `sortedScrollableAncestors`. I've since refactored this approach to only give consumers the option decide whether to set the scroll order to the tree order or reversed tree order https://github.com/clauderic/dnd-kit/pull/140/commits/800b2d630e4ad0987bb31d7224bbecff1ac769b5

mitchheddles: What if you accept a sort function instead?

mitchheddles: Semi related to this change... is (or will) there be a way to stop an element scrolling?
When I first saw this PR I was thinking I could use it to stop the window scrolling horizontally, which is an issue I'm facing now with a vertical-only scrolling list.

clauderic: > What if you accept a sort function instead?

I'd prefer to limit the use-cases here for now, can't really think of good use-cases where you would want to scroll ancestors in an order that doesn't match the DOM (reversed or otherwise).

>Semi related to this change... is (or will) there be a way to stop an element scrolling?

I hadn't considered that use-case, but that could certainly be added.

Scroll ancestors are necessary in a few places in the library for collision detection to work properly, so I'd prefer to only let consumers decide if a given ancestor can be scrolled.

Something like:

```tsx
// canScroll(scrollAncestor: Element): boolean;

<DndContext autoScroll={{
  enabled: true,
  scrollOrder: ScrollOrder.TreeOrder,
  canScroll(element) {
    if (element === document.scrollingElement) {
      return false;
    }

    return true;
  }
}} />
```

mitchheddles: This looks good to me and I agree on the sorting 👍

rsshilli:
@andelkocvjetkovic Thank you. You saved my sanity. @clauderic Maybe you could consider editing your first comment above?

PR #144, opened 2021-03-21
Allow consumers to determine when to perform layout animations and when to measure nodes

clauderic:
This PR allows consumers to customize or override the internal heuristics for determining whether derived transformations should be applied to allow a sortable node to perform layout animations.

By default, layout animations via derived transforms happen when sorting is cancelled or when new elements are added / removed within the list during sorting.

Letting consumers override this option opens up a lot of new possibilities for consumers, such as easily animating adding / deleting items (see https://github.com/clauderic/dnd-kit/discussions/108) or only animating items to their new positions on drop rather than while dragging:

https://user-images.githubusercontent.com/1416436/111924662-06d2fb00-8a7c-11eb-8ec3-46981a58fe5f.mp4

https://user-images.githubusercontent.com/1416436/112006140-d29e1f80-8af9-11eb-88e4-0edb953924d9.mp4

https://user-images.githubusercontent.com/1416436/112242616-987c6d00-8c22-11eb-82c4-2aefa4adf04d.mp4


GNUGradyn:
Can this be added to the docs?

PR #171, opened 2021-03-29
Add more flexibility for distance and tolerance measurements

mitchheddles:
This PR adds a new `DistanceMeasurement` option to the preexisting `DistanceConstraint` and `DelayConstraint` constraints.

```ts
export type DistanceMeasurement =
  | number
  | Coordinates
  | CoordinatesX
  | CoordinatesY;

interface DistanceConstraint {
  distance: DistanceMeasurement;
}

interface DelayConstraint {
  delay: number;
  tolerance: DistanceMeasurement;
}
```

The goal of this option is to let a user specify which axis, `x`, `y` or both (`xy`) the activation `distance`, or `tolerance` should be measured on.

This option is useful for drag and drop contexts that are restricted to a single axis.

**Example usage:**
```ts
const sensors = useSensors(
  useSensor(PointerSensor, {
    activationConstraint: {
      distance: { y: 7 },
    },
  }),
  useSensor(KeyboardSensor, {
    coordinateGetter: sortableKeyboardCoordinates,
  }),
);
```

An alternate solution could be to allow `distance` as an object, e.g. `distance: { x?: number, y?: number }`
But it's up to you 🙂


mitchheddles:
@clauderic I haven't been able to test this locally, I ran into hook issues when linking the package.

Please let me know if you have any trouble or concerns with the change. It would be great if we could get this, or something similar working.

packages/core/src/sensors/pointer/AbstractPointerSensor.ts
```diff
@@ -9,11 +14,13 @@ import type {Coordinates} from '../../types';
...
 interface DistanceConstraint {
   distance: number;
+  axis?: Axis;
 }
 
 interface DelayConstraint {
   delay: number;
   tolerance: number;
+  axis?: Axis;
 }
```
clauderic, line 24: What if both `distance` and `tolerance` accepted either a number or `x` and `y` coordinates instead?

```suggestion
interface DistanceConstraint {
  distance: number | Coordinates;
}

interface DelayConstraint {
  delay: number;
  tolerance: number | Coordinates;
}
```

I could see a use-case for consumers wanting to have different values for the `x` and `y` axis.

In your specific scenario, a consumer would be able to set `distance` to `{x: Infinity, y: 10}` for example

clauderic (commented):
Thanks for your contribution! I have a few suggestions about the implementation, but otherwise I agree that this would be a valuable addition.

If you have time, would you be willing to add Cypress integration tests for this? You can take a look at these tests for reference:

https://github.com/clauderic/dnd-kit/blob/master/cypress/integration/draggable_spec.ts#L188-L218

packages/core/src/sensors/pointer/AbstractPointerSensor.ts
clauderic: Thinking about this more, in your scenario if I'm understanding correctly, you only care about distance on the `y` axis and not on the `x` axis, so perhaps accepting `number | {x: number} | {y: number} | {x: number, y: number}` makes the most sense

mitchheddles: That's correct.

With the proposed options we now need to decide how we handle `{x: number, y: number}`... is it `x > dx && y > dy` or `x > dx || y > dy`.

clauderic: Good question. The ideal API to represent this is likely something like:

```ts
{x: 10} // x >= dx
{y: 20} // y >= dy
{x: 10, y: 20} // x >= dx && y >= dy
[{x: 10}, {y: 20}] // x >= dx || y >= dy
```
I don't want to scope creep on your PR however, so it may not be the end of the world if we don't support the `x >= dx || y >= dy` syntax initially.

mitchheddles: To be honest, I think `x >= dx || y >= dy` is covered by `number` which will be the absolute distance from the origin... which is better anyway.

clauderic: Agreed

mitchheddles: @clauderic I've addressed this in https://github.com/clauderic/dnd-kit/pull/171/commits/e61344d189282928cdad4eebc6fcc6e28deafe12

packages/core/src/sensors/utilities/hasExceededDistance.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export function hasExceededDistance(
+  delta: Coordinates,
+  measurement: DistanceMeasurement
+): boolean {
+  const dx = Math.abs(delta.x);
+  const dy = Math.abs(delta.y);
+
+  if (typeof measurement === 'number') {
+    return Math.sqrt(dx ** 2 + dy ** 2) > measurement;
```
clauderic, line 11: Nice, this was an oversight in the original implementation 👍

clauderic (approved):

mitchheddles:
@clauderic Thanks for jumping into the PR, I've added some test cases in https://github.com/clauderic/dnd-kit/pull/171/commits/c60600011205073c051eaaa01c3154299b73c799

clauderic:
This has shipped in version 2.1.0 of `@dnd-kit/core`, thanks again for your contribution

mitchheddles:
Thanks for the quick turn around @clauderic!

PR #177, opened 2021-04-01
Fix auto scroll top and left speed calculations

FinestV:
Auto scroll speed for top and left is currently calculated incorrectly. It scrolls very very fast in these directions depending on the scrollable containers offset to the edge of the page.

The Vertical scrollable stories currently mask the issue because the scrollable containers sit at the very top of the page. This PR also updates the stories.


clauderic (approved):
Thanks for submitting a PR ❤️ This was definitely an oversight, your changes looks good.

Issue #139, opened 2021-03-18
Is it possible to add more data about the draggable in DragStartEvent?


jaymathew:
Hey, Thanks so much for this library. 

Is it possible to add more data into the DragStartEvent? I know you can see the id currently. I would need to filter through an extensive list to find the object I need, and using an index as the id wouldn't work as I render two arrays in one DndContext. 

Thank you,

salali:
Would it be possible for the useDroppable `data` property to be passed to the `DndContext` event handlers?

interface UseDroppableArguments {
  id: string;
  disabled?: boolean;
  data?: Record<string, any>; // Add this property to the event.active object?
}

clauderic:
Adding `data` to draggable items and exposing `data` of the `over` droppable in handlers is planned, just haven't had time to get around to it.

@salali if you want to get started working on a PR with that change I'd gladly work with you to get it done.

norbertas-gaulia:
That would be great addition. 
Is there any way to get any info about draggable currently? If i want to switch css class based on who is over? Thanks

clauderic:
@norbertas-gaulia you can already get info about the current draggable and what droppable it is over in what is returned by `useDraggable()` and `useDroppable` or in the `onDragOver` callbacks of `DndContext`.

This is documented in the [Quick start guide](https://docs.dndkit.com/introduction/getting-started) along with the docs for [`useDraggable`](https://docs.dndkit.com/api-documentation/draggable/usedraggable#over) and [`useDroppable`](https://docs.dndkit.com/api-documentation/droppable/usedroppable#over).

norbertas-gaulia:
Correct, i missed `over` prop from `useDroppable `. Thanks.

clauderic:
> But it gives droppable node and droppable id, what if i want draggable id in droppable?

```tsx
import {useDndContext} from '@dnd-kit/core';

function Droppable() {
  // You can get the `id` of the active draggable 
  const {active} = useDndContext();
}
```

growthwp:
[...]

PR #207, opened 2021-04-16
Normalize and expose data for draggable and droppable nodes

clauderic:
The `data` argument for `useDraggable` and `useDroppable` is now exposed in event handlers and on the `active` and `over` objects.

**Example usage:**

```tsx
import {DndContext, useDraggable, useDroppable} from '@dnd-kit/core';

function Draggable() {
  const { attributes, listeners, setNodeRef, transform } = useDraggable({
    id: 'draggable',
    data: {
      type: 'type1',
    },
  });

  /* ... */
};

function Droppable() {
  const { setNodeRef } = useDroppable({
    id: 'droppable',
    data: {
      accepts: ['type1', 'type2'],
    },
  });
  
  /* ... */
};

function App() {
  return (
    <DndContext onDragEnd={({active, over}) => {
      if (over?.data.current.accepts.includes(active.data.current.type)) {
        // do stuff
      }
    }} />
  );
}
```

### DIFF
--- a/.changeset/expose-data-ref.md
+++ b/.changeset/expose-data-ref.md
@@ -1,0 +1,46 @@
+---
+'@dnd-kit/core': minor
+'@dnd-kit/sortable': minor
+---
+
+The `data` argument for `useDraggable` and `useDroppable` is now exposed in event handlers and on the `active` and `over` objects.
+
+**Example usage:**
+
+```tsx
+import {DndContext, useDraggable, useDroppable} from '@dnd-kit/core';
+
+function Draggable() {
+  const {attributes, listeners, setNodeRef, transform} = useDraggable({
+    id: 'draggable',
+    data: {
+      type: 'type1',
+    },
+  });
+
+  /* ... */
+}
+
+function Droppable() {
+  const {setNodeRef} = useDroppable({
+    id: 'droppable',
+    data: {
+      accepts: ['type1', 'type2'],
+    },
+  });
+
+  /* ... */
+}
+
+function App() {
+  return (
+    <DndContext
+      onDragEnd={({active, over}) => {
+        if (over?.data.current.accepts.includes(active.data.current.type)) {
+          // do stuff
+        }
+      }}
+    />
+  );
+}
+```
--- a/packages/core/src/components/DndContext/DndContext.tsx
+++ b/packages/core/src/components/DndContext/DndContext.tsx
@@ -23,7 +23,7 @@ import {
   getInitialState,
   reducer,
 } from '../../store';
-import type {Coordinates, ViewRect, LayoutRect} from '../../types';
+import type {ViewRect} from '../../types';
 import {DndMonitorContext, DndMonitorState} from '../../hooks/monitor';
 import {
   useAutoScroller,
@@ -61,10 +61,10 @@ import {
 } from '../../utilities';
 import {applyModifiers, Modifiers} from '../../modifiers';
 import type {
-  LayoutRectMap,
-  DraggableNode,
-  DraggableNodes,
+  Active,
   DroppableContainers,
+  DroppableContainer,
+  DataRef,
 } from '../../store/types';
 import type {
   DragStartEvent,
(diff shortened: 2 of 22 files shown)


growthwp:
Yo!

Thanks for the speedy work, it should allow somewhat more complex interactions to happen, can't wait, but I think it's important to keep one thing in mind and remember why I/others have even asked about this: we wanna freely make draggables interact with droppables **regardless of what context they're in.**. Sorry if what I'm about to write seems overly detailed but the nuances really matter:

As it stands right now, `data` in these hooks or not, we're still under the same context, meaning, if we were to have a set of draggables interact with a droppable not in the same context, well, that can't happen. The current model implies that each context handles its drag events (`dragEnd, dragStart`, etc.), collisions, sensors individually. With the new model, the system would allow *any* draggable to be dropped into any droppable, then, it's up to some callbacks (be them just the `accepts` key or something more advanced) to decide if a droppable can indeed work with a draggable.

I just can't find a good reason not to operate in a manner where, as you pointed out in: https://github.com/clauderic/dnd-kit/issues/139#issuecomment-814178995 , every droppable decides how the collisions happen (better yet, let a central figure handle it, but I guess that's more advanced), you don't have coupling between draggables/droppables besides the newly introduced `accepts` argument. If we approach things this way, then there are basically no limits to what you can build.

So, then I thought "well, I really appreciate the react way of doing things, wrapping components and so is elegant" and that's nice, but take a look at the model of the [createSlice method from Redux](https://redux-toolkit.js.org/api/createSlice#createslice), which basically lets you create a small portion of data that you care about within the store - with the store being unique, declared at the top of your application. This means that all data flow is visible and you can, through `useSelector`, declare what you care about.

This flux-like model gave me inspiration - why are we locking ourselves like this? Assuming we can create a top-level, unique context, our codebase would end up looking something like this:
```js
const Droppable = () => {
  const { attributes, listeners, setNodeRef, transform } = useDraggable({
    id: 'uniqueId',
    types: ['type1', 'type2'],
    data: {},
  });
};

const Draggable = () => {
  const { setNodeRef } = useDroppable({
    id: 'uniqueId',
    accepts: {
      type1: {
        collision: customAlgo,
        sensors: {}
      },
      type2: {
        collision: customAlgo2,
      }
    },
  });
};
```
And if we were to have a [Redux Dev Tools-like thingie](https://www.youtube.com/watch?v=tHLRexgkOzk), whenever we'd have a drag, drop and any other event, everyone could see it **but not everyone should react to it.** If we were to re-implemen the `useSortable` hook, we'd simply create a "slice" inside the context that the sortable can use (so, while everything is visible, it's still encapsulated - extremely important point) and it does its own thing while we can monitor/change it from the outside.

Anyways, sorry if this got a bit longer than expected, but this was my motivation and, I assume, based on what I saw being built with DnD libraries, many others'. There has to *eventually* be a way to see data flow like this without needing to hack something ugly in.

clauderic:
> we wanna freely make draggables interact with droppables regardless of what context they're in

Having draggable and droppable elements from **different** `DndContext` providers interact with one another will never happen. There are dozens of reasons why that could never work.

I think you're approaching this problem with the wrong mental model. `DndContext` is your central store. The goal is to get to a place where it's easier to manage multiple types of drag and drop interactions under the same central `DndContext` provider, not to make it so that you can have elements under different `DndContext` providers interact with one another.

growthwp:
Aye! I've been reading the source for a week now and will hopefully understand it as soon as possible so I can start contributing and helping, but I think we're talking about the same thing and I just went a bit too "broad" and bringing in Redux as a philosophy. Perhaps my wording made it seem like the "free flow" of things wouldn't be in a central store, which you already have.

Perhaps a better wording would be "operating within the same context but being able to discern between interactions", which seems to be the goal here.

An example of how having two contexts mess things up is when you have a `SortableContext` inside a `DnDContext`, without the `SortableContext`, a draggable is able to see an actual sortable item as a draggable as well, so, my thinking was "well, that `SortableContext` could just insert itself into the bigger context but obviously manage its slice of state only and not care about others from the outside". Internally, each `DnDContext` manages itself in a sandbox way, it's just that, since they are all descendants of a bigger context, they could still be aware of each other's presence.

I've seen the `useDndMonitor` hook and I already have an idea on how to tie things together. Eagerly waiting on this update to see what I can build with it.

PR #224, opened 2021-04-21
Fix #223 Rects on the active property should be typed as ViewRect instead of LayoutRect

### LINKED ISSUE #223: Rects on the `active` property should be typed as `ViewRect` instead of `LayoutRect`
When I use `console.log` to print `event.active.rect.current.initial`, there are `bottom, height, left, offsetLeft, offsetTop, right, top, width` 8 attributes. But in typescript its type `LayoutRect` has only `width, height, offsetLeft, offsetTop` 4 properties. So it's impossible to access the rest of properties without converting it to `any` first.

dzcpy:
#223

### DIFF
--- a/packages/core/src/store/types.ts
+++ b/packages/core/src/store/types.ts
@@ -4,7 +4,6 @@ import type {
   Coordinates,
   ViewRect,
   ClientRect,
-  LayoutRect,
   UniqueIdentifier,
 } from '../types';
 import type {SyntheticListeners} from '../hooks/utilities';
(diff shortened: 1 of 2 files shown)


packages/core/src/store/types.ts
```diff
@@ -34,14 +33,14 @@ export interface Active {
   id: UniqueIdentifier;
   data: DataRef;
   rect: MutableRefObject<{
-    initial: LayoutRect | null;
-    translated: LayoutRect | null;
+    initial: ViewRect | null;
+    translated: ViewRect | null;
```
clauderic, line 37: The scope of your PR should only be to change these two lines

clauderic (changes requested):

packages/core/src/store/types.ts
dzcpy: I tried that but there are failed tests. I'm not quite familia with the code structure and each variables function. Maybe I'll just close this PR it doesn't seem to work

PR #228, opened 2021-04-22
Enable Customize trigger function for Sensors

Sannma7:


clauderic:
Hey @Sannma7, 

I'm a bit hesitant to merge this just because it's fairly straightforward for consumers to override the default activator of a sensor.

```ts
class MyPointerSensor extends PointerSensor {
  static activators = [
    {
      eventName: 'onPointerDown' as const,
      handler: (
        {nativeEvent: event}: PointerEvent,
        options: PointerSensorOptions
      ) => {
        return true;
      },
    },
  ]
}
```

I'm also unclear how well this approach would scale for consumers if there were multiple activators defined for a given sensor.

Sannma7:
I did override the activator in this approach to make my sensor only triggered by a pen.
I opened this pr because I think it's more friendly to consumers to do this without having to dig into source code.

What about just export the `AbstractPointerSensor`?

clauderic:

PR #240, opened 2021-04-29
fix: fix wrong position with transform style

### LINKED ISSUE #205: Droppable - Wrong Droppable area with transform style
Wrong Droppable area when using transfrom style on Droppable container, how can i fix it?

demo: https://codesandbox.io/s/stoic-river-iwsl3?file=/src/App.js

a1ooha:
Fix #205

### DIFF
--- a/.changeset/lazy-pillows-eat.md
+++ b/.changeset/lazy-pillows-eat.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/core': patch
+---
+
+using `getBoundingClientRect` to get right position with transform style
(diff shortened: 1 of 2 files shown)


clauderic:
Hey @a1ooha,

Thanks for opening a PR, but unfortunately you'll have to go back to the drawing board as there are very good reasons why the library is built this way.

Transforms need to be ignored for all use-cases that use the `sortable` preset. You can see what happens if you tophat your changes with the sortable stories. The problem isn't initially visible because items aren't transformed on drag start, but if you have items that are dynamically measured, you'll see the issues right away:

https://user-images.githubusercontent.com/1416436/116630919-30f0c600-a922-11eb-8fd7-7514cfb99be4.mp4


One potential solution however would be to let consumers override how to measure `LayoutRect`s via a config prop on `DndContext`.

PR #243, opened 2021-04-30
Handle overlay overflow as a scrollable element

py-wai:
Update regex used in `isScrollable`, to consider element with `overflow: overlay` as a scrollable element.


clauderic:
Thanks!

PR #246, opened 2021-05-01
Fix style prop of DragOverlay to work

inokawa:
As written [here in docs](https://docs.dndkit.com/api-documentation/draggable/drag-overlay#class-name-and-inline-styles), it looks like that the style prop of DragOverlay should be passed to the wrapper element, but not passed now.

### DIFF
--- a/packages/core/src/components/DragOverlay/DragOverlay.tsx
+++ b/packages/core/src/components/DragOverlay/DragOverlay.tsx
@@ -40,7 +40,7 @@ export const DragOverlay = React.memo(
     adjustScale = false,
     children,
     dropAnimation = defaultDropAnimation,
-    style: propStyle,
+    style: styleProp,
     transition = defaultTransition,
     modifiers,
     wrapperElement = 'div',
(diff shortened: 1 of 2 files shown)


packages/core/src/components/DragOverlay/DragOverlay.tsx
```diff
@@ -88,6 +89,7 @@ export const DragOverlay = React.memo(
         };
     const style: React.CSSProperties | undefined = activeNodeRect
       ? {
+          ...propStyle,
```
arpithindukuri, line 92: I think it would be better to move `...propStyle,` down to the end of `style`.

If we are giving users the ability to control the style on such a low level, I think they should have full control of the prop, so they can override styles if needed.

Just my two cents. Thank you @inokawa for taking the initiative!! I had this problem and appreciate you fixing this.

inokawa: I agree with you. Thank you!
I fixed in https://github.com/clauderic/dnd-kit/pull/246/commits/b9b0f73e02eb7ee32f7da15ff90548dbd1878efc

Issue #293, opened 2021-05-26
sorting does not work with yarn 2


akhmadullin:
Hi, first of all, thx dnd-kit team for a such great lib!

### Problem

We tried to use dnd-kit for making sortable gallery in our project, based on yarn 2.
We faced a problem - @dnd-kit/sortable does not work in our project, despite the fact we did everything like in the docs.

**Example repo**
https://github.com/akhmadullin/dnd-kit-demo-yarn-2-broken
**Live demo**
https://akhmadullin.github.io/dnd-kit-demo-yarn-2-broken/

With yarn 1 everything is ok - sorting works well.

**Example repo**
https://github.com/akhmadullin/dnd-kit-demo-yarn-1
**Live demo**
https://akhmadullin.github.io/dnd-kit-demo-yarn-1/

**Attempt to fix bug in yarn 2 based project**

Debugging shows, that the reason of this bug is different DndContext in useDraggable and useDroppable hooks (different from App DndContext). This happens because of Yarn 2 dependencies resolving mechanism - we have 2 different @dnd-kit/core instances in our App.

**Solution for projects, that use @dnd-kit/core and @dnd-kit/sortable in dependencies**
To solve problem with dnd-kit dependencies you should define in .yarnrc.yml, that @dnd-kit/core should be used as peerDependency for @dnd-kit/sortable:

```yml
# .yarnrc.yml
packageExtensions:
  "@dnd-kit/sortable@*":
    peerDependencies:
      "@dnd-kit/core": "*"
```

**Example repo**
https://github.com/akhmadullin/dnd-kit-demo-yarn-2-fixed
**Live demo**
https://akhmadullin.github.io/dnd-kit-demo-yarn-2-fixed/

### Possible solution

It seems like, dnd-kit packages should use each other as peerDependency inside of dnd-kit project.

PR #295, opened 2021-05-26
Move @dnd-kit/core as a `peerDependencies` for other @dnd-kit packages that depend on it

akhmadullin:
solves #293


clauderic:
I believe only `@dnd-kit/core` should be listed as a peerDependency of the other packages, otherwise consumers will need to manually specify packages such as `@dnd-kit/accessibility` in their dependencies, which isn't something they should have to do.

akhmadullin:
Ok

So, there will also be no problem with `@dnd-kit/utilities`, if it stays mentioned as dependency?

Now `@dnd-kit/utilities` is used as dependency in `@dnd-kit/core` and also it mentioned as dependency in `@dnd-kit/modifiers` and `@dnd-kit/sortable`, which, in turn, use `@dnd-kit/core` as dependency.

How do you think, is it possible, that someone can face with problem of having several different versions of  `@dnd-kit/utilities` in his project, like we had 2 different versions of `@dnd-kit/core`?

If it is not possible, I will list only `@dnd-kit/core` as peerDependency.

Is it ok, if I just edit my previous commits and force-push them or I should push new commits, based on my previous ones, and they will be merged/squashed soon all together?

clauderic:
It could potentially happen for consumers to end up with two versions of `@dnd-kit/utilities`, but it's less problematic than having two versions of `@dnd-kit/core`. Having two versions of `@dnd-kit/core` causes issues with React context which causes the issues of `@dnd-kit/sortable` not working at all.

`@dnd-kit/utilities` is a package that does not change often, and shouldn't have breaking changes that would cause major issues even if two versions were installed. Installing `@dnd-kit/utilities` should be optional for consumers, which is why I'm reluctant for it to be added as a peer dependency.

akhmadullin:
Ok, I got it

akhmadullin:
Is it ok, if I just force-push edited commits instead of current commits?

clauderic:
Yes that's fine, just make sure you update the `.changeset` to reflect the new changes

akhmadullin:
@clauderic, I updated commits.

It seems, that they are ok now.

Should I do anything else to get this PR approved?

Issue #291, opened 2021-05-21
Showing stories code in storybook


adelin-b:
Hello, today the storybook is only usefull as documentation by checking the corresponding repo on github and finding the related stories file. 

Showing the content of thoses stories next to the component using an addon would be usefull.

ddublon:
As well as here, it didnt solved the problem @clauderic @akhmadullin @adelin-b

clauderic:
This issue was opened against a previous version of dnd-kit. The library has since undergone a major rewrite.

It has been rebuilt from the ground up with support for **multiple frameworks** (React, Vue, Svelte, Solid, and vanilla JavaScript), improved performance, and a new API.

- **Documentation**: https://dndkit.com
- **Migration guide**: https://dndkit.com/react/guides/migration
- **GitHub**: https://github.com/clauderic/dnd-kit

If this issue is still relevant to the latest version, please open a new issue.

PR #301, opened 2021-05-28
Add @storybook/addon-storysource

akhmadullin:
solves https://github.com/clauderic/dnd-kit/issues/291

This PR adds @storybook/addon-storysource to the dnd-kit project. This addon provides each story page with its source code.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -6,6 +6,7 @@ module.exports = {
     '@storybook/addon-essentials',
     '@storybook/addon-actions',
     '@storybook/addon-links',
+    '@storybook/addon-storysource',
   ],
   webpackFinal: async (config) => {
     config.module.rules = [
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "@storybook/addon-essentials": "^6.1.11",
     "@storybook/addon-info": "^5.3.21",
     "@storybook/addon-links": "^6.1.11",
+    "@storybook/addon-storysource": "^6.1.11",
     "@storybook/addons": "^6.1.11",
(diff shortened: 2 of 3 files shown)


akhmadullin:
I'm not sure about necessity of `changeset` for this PR, because all dnd-kit packages stayed the same.
I just changed some files in root repo.

Should I add ``changeset`` in this case?

If yes, what dnd-kit packages to select via ``yarn .changeset``?

clauderic:
Hey @akhmadullin, thanks for your contribution. I'm not sure if merging this PR would really address the issue described in #291.

In practice, most stories are authored for code shareability and are hard to read. Having this addon enabled wouldn't really solve that.

The code it would show would look something like this, which isn't super helpful in my opinion:

```tsx
export const BasicSetup = () => <Sortable {...props} />;

export const WithoutDragOverlay = () => (
  <Sortable {...props} useDragOverlay={false} itemCount={5} />
);

export const DragHandle = () => <Sortable {...props} handle />;
```

akhmadullin:
@clauderic, ok

I also expected, that this addon would add the source code of whole ``.story.tsx`` file - previously addon had worked in this way, but now it shows only the source code of selected story.

[Addon docs](https://www.npmjs.com/package/@storybook/addon-storysource) says, that the addon behavior changed in 6.0.0 version, and it has options to enable full source code (of whole file) - I tried to use it, but it does not work.

I will try to fix it, but I'm not sure, that my attempt will be successful.

ddublon:
can you make it more easy to use please?
why it is so compilated to see the code?
i want to copy and past the example and then modify it to my needs

PR #305, opened 2021-05-29
Fix flickering of node in overlay with React concurrent mode

### LINKED ISSUE #304: Flickering with DragOverlay after drop animation finished
When using `DragOverlay` with React in concurrent mode, the dragged item inside the overlay just shows up again for some milliseconds, where it was released after the drop animation. This screencast demonstrates this problem:

dnd-kit-flickering

**I am absolutely aware that concurrent mode is experimental**, but I think it justs demasks an already existing problem. Because after the [drop animation finished](https://github.com/clauderic/dnd-kit/blob/ae398de012aee28f5e3bec10b438153d00f65630/packages/core/src/components/DragOverlay/hooks/useDropAnimation.ts#L97), it seems that the dragged node in the overlay gets its previous transform back (where it was released from). In React legacy mode, it doesn't matter as the DOM is not updated fast enough, but in concurrent mode, the node in the overlay shows up for some milliseconds where it was released before [the overlay itself is removed](https://github.com/clauderic/dnd-kit/blob/ae398de012aee28f5e3bec10b438153d00f65630/packages/core/src/components/DragOverlay/DragOverlay.tsx#L160).

A possible fix is to hide the dragged node in the overlay after the drag animation is finished. It will be recreated on the next drag anyway. I'll create a PR if you are ok with this.

medihack:
Fixes #304

### DIFF
--- a/.changeset/two-terms-mate.md
+++ b/.changeset/two-terms-mate.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/core': patch
+---
+
+Hide the node in the overlay after the drop animation is finished. This prevents some flickering with React concurrent mode.
--- a/packages/core/src/components/DragOverlay/hooks/useDropAnimation.ts
+++ b/packages/core/src/components/DragOverlay/hooks/useDropAnimation.ts
@@ -95,6 +95,8 @@ export function useDropAnimation({
             );
 
             nodeAnimation.onfinish = () => {
+              node.style.display = 'none';
+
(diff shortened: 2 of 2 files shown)


clauderic (approved):
Good catch, thanks for reporting this and opening a PR 🙌

medihack:
I have to thank you for this very flexible library. And sorry, I wasn't quite aware of how the changesets exactly work. So you have to rename this one also. Next time I know!

PR #316, opened 2021-06-04
Add optional onDragMove screen reader announcements

lsit:
Sometimes you need the screen reader to announce changes that happen `onDragMove`, not `onDragOver`, such as the nesting in the `SortableTree` example.

I've made this announcement optional and left it out of the `defaultAnnouncements` since it probably won't be used very often. I've also updated the `SortableTree` example to make use of it.

### DIFF
--- a/stories/3 - Examples/Tree/SortableTree.tsx
+++ b/stories/3 - Examples/Tree/SortableTree.tsx
@@ -93,6 +93,11 @@ export function SortableTree({
   const [activeId, setActiveId] = useState<string | null>(null);
   const [overId, setOverId] = useState<string | null>(null);
   const [offsetLeft, setOffsetLeft] = useState(0);
+  const [currentPosition, setCurrentPosition] = useState<{
+    parentId: string | null;
+    overId: string;
+  } | null>(null);
+
   const flattenedItems = useMemo(() => {
     const flattenedTree = flattenTree(items);
     const collapsedItems = flattenedTree.reduce<string[]>(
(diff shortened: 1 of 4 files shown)


stories/3 - Examples/Tree/SortableTree.tsx
```diff
@@ -143,8 +144,39 @@ export function SortableTree({
...
+      if (overId) {
+        return `${id} was moved over ${overId}.`;
+      }
+
+      return;
+    },
+    onDragEnd(id, overId) {
+      if (overId) {
+        return `${id} was dropped over ${overId}`;
+      }
```
clauderic, line 168: Hey @lsit, thanks for adding an example of how this should be used. What's the distinction between `over` and `under`? Are there cases where we announce `over` but should be announcing `under` (and vice-versa)?

I'd imagine if VoiceOver is on but you're dragging with the mouse that the announcements may announce both `onDragMove` and  `onDragOver` in quick succession 🤔

lsit: I've updated the announcements to be more accurate for when to use `moved` versus `nested`. `nested` is used when the item being moved has greater depth than the item it is over. `moved` is used for all the other cases.

VoiceOver only reads out the last string that is given to the live region so you won't hear multiple announcements in quick succession when dragging with the mouse. Besides that, I've updated it so only one of the events will generate a new announcement to be read at any one time.

stories/3 - Examples/Tree/SortableTree.tsx
```diff
@@ -143,8 +149,38 @@ export function SortableTree({
...
   }, [flattenedItems, offsetLeft]);
 
+  const announcements: Announcements = {
+    onDragStart(id) {
+      if (projected) {
+        setCurrentPosition({
+          parentId: projected.parentId,
+          overId: id,
+        });
+      }
```
clauderic, line 159: Can this be moved to `handleDragStart`? Announcements should be pure and not have side-effects like updating internal state.

This does strike me as something that could potentially be derived from some other state (such as `active`), but not the end of the world to have a separate state property for it.

stories/3 - Examples/Tree/SortableTree.tsx
```diff
@@ -143,8 +149,38 @@ export function SortableTree({
...
+      return `Picked up ${id}.`;
+    },
+    onDragMove(id, overId) {
+      return getMovementAnnouncement('onDragMove', id, overId);
+    },
+    onDragOver(id, overId) {
+      return getMovementAnnouncement('onDragOver', id, overId);
+    },
+    onDragEnd(id, overId) {
+      setCurrentPosition(null);
```
clauderic, line 170: Same comment as above about announcements being pure, can we move this to `handleDragEnd`?

stories/3 - Examples/Tree/SortableTree.tsx
```diff
@@ -143,8 +149,38 @@ export function SortableTree({
...
+
+      return getMovementAnnouncement('onDragEnd', id, overId);
+    },
+    onDragCancel(id) {
+      setCurrentPosition(null);
```
clauderic, line 175: Same comment as above about announcements being pure, can we move this to `handleDragCancel`?

clauderic (commented):
Looks great, thanks for updating the announcements!

Just a couple final comments

clauderic (approved):
Looks great! Thanks again for contributing and updating the example usage 💯

Issue #172, opened 2021-03-29
Sortable - onClick event fired when moving a SortableItem to a lower index 


lonk:
Hi !

I'm facing an issue using sortable: when I'm moving a SortableItem from right to left, the onClick event is fired along the onDragEnd event. This isn't happing if I'm moving my item to the right.

Here is the simple sample of code I'm using:
``` 
              <DndContext
                    sensors={sensors}
                    collisionDetection={closestCenter}
                    onDragEnd={handleDragEnd}
                    modifiers={[restrictToParentElement]}
                >
                    <SortableContext
                        items={displayedArticles.map(article => article.id)}
                        strategy={rectSortingStrategy}
                    >
                        {displayedArticles.map(article => (
                            <Article
                                key={article.id}
                                name={article.name}
                                id={article.id}
                                onDeleteClick={() => {
                                    setArticleToDelete(article.id);
                                }}
                                onClick={() => console.log('onClick fired')}
                            />
                        ))}
                        <Add onClick={onOpen} />
                    </SortableContext>
                </DndContext>
```

And here is a video of what's happening:
![ezgif com-gif-maker (1)](https://user-images.githubusercontent.com/2729398/112909698-3cb65600-90f2-11eb-9332-0d99237a5aee.gif)

Is this an expected behavior ? Do I miss something ?
Is there a way to completely silent onClick when the Item has been dragged ?

Thank you !

lonk:
Hi @clauderic !

Thank you for your quick reply.

Your workaround doesn't work in my case because onDragEnd is fired before onClick so I think I'll go with the DragOverlay workaround :)

Have a nice day !

tjunghans:
I have a similar issue with items in a grid, where the first item (cell) is larger than the rest. When dragging any item to the first  cell (top, left), the click event is fired. For any other re-order, the click event is not fired (as expected). This includes dragging the first cell to another position, which also works as expected.

I've tried the above suggestions without success. The click event is fired after dragEnd and DragOverlay creates new problems for me. I have one last idea and that is to track when the first cell is being replaced and ignore the first click after drag start.

firede:
> onClick event fired when moving a SortableItem to a lower index

Got the same issue, I created a CodeSandbox: <https://codesandbox.io/s/sortabletabs-without-dragoverlay-vg3vg?file=/src/SortableTabs.tsx>

---

[...]

PR #377, opened 2021-07-22
Stop propagation of click events on draggable node once activation constraints are met

clauderic:
Fixes https://github.com/clauderic/dnd-kit/issues/172

This PR adds a `click` event listener to the draggable node to stop propagation of click events once activation constraints are met.

If activation constraints are not met, `click` events fire as they should.

### DIFF
--- a/.changeset/prevent-click-propagation.md
+++ b/.changeset/prevent-click-propagation.md
@@ -1,0 +1,5 @@
+---
+"@dnd-kit/core": minor
+---
+
+Pointer, Mouse and Touch sensors now stop propagation of click events on the draggable node once activation constraints are met
--- a/packages/core/src/sensors/events.ts
+++ b/packages/core/src/sensors/events.ts
@@ -1,6 +1,15 @@
 export enum EventName {
+  Click = 'click',
   Keydown = 'keydown',
   ContextMenu = 'contextmenu',
   Resize = 'resize',
   VisibilityChange = 'visibilitychange',
+}
+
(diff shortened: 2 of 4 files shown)


andys8:
Hi @clauderic, I'm working on application where cards can be sorted. These cards contain checkboxes and buttons. With this change (since `4.0.0`) this will swallow these all click events and these buttons don't work anymore.

Is there a way to these buttons can still work?

HelKyle:
Hi @andys8, to avoid the conflicts between click and drag events, you can implement a custom MouseSensor like this.

```js
import type { MouseEvent, KeyboardEvent } from 'react'
import {
  MouseSensor as LibMouseSensor,
  KeyboardSensor as LibKeyboardSensor
} from '@dnd-kit/core'

export class MouseSensor extends LibMouseSensor {
  static activators = [
    {
      eventName: 'onMouseDown' as const,
      handler: ({ nativeEvent: event }: MouseEvent) => {
        return shouldHandleEvent(event.target as HTMLElement)
      }
    }
  ]
}

function shouldHandleEvent(element: HTMLElement | null) {
  let cur = element

  while (cur) {
    if (cur.dataset && cur.dataset.noDnd) {
      return false
    }
    cur = cur.parentElement
  }

  return true
}
```

RemyMachado:
@HelKyle 
It works like a charm, thanks!

I prefer this implementation that doesn't relay on `activation constraints`.
This code snippet deserves a place in the official documentation. The library is great, but the lack of examples is a bummer.

PR #399, opened 2021-08-04
Support for tolerance in DistanceConstraint

supersebh:


clauderic (commented):
This looks great! Would you mind adding stories and integration tests for minimum distance y and tolerance x?

I'm also a bit unclear as to how this should work if you provided a tolerance that was smaller than the distance or a tolerance that was in a different axis than the distance 🤔

supersebh:
@clauderic Thank you for the reply!

> This looks great! Would you mind adding stories and integration tests for minimum distance y and tolerance x?

Done!

> I'm also a bit unclear as to how this should work if you provided a tolerance that was smaller than the distance or a tolerance that was in a different axis than the distance 🤔

The way it's working right now is as soon as you exceed the `tolerance` the drag operation will be aborted, unless it has already started (since `distance` criteria was met). So you need to be careful and not pick conflicting settings for `distance` and `tolerance` if used together. E.g.

```typescript
{
  distance: {x: 10},
  tolerance: {x: 5},
}
```

won't make sense since you can never meet the `distance` criteria without meeting the `tolerance` criteria first.

Tolerance in different axis than the distance is exactly how we're using it right now in our project, and that is described [here](https://github.com/clauderic/dnd-kit/issues/400#issue-960481720).

Please let me know if anything is unclear!

clauderic:
Sounds good 👍  Could you [add a changeset to your PR](https://github.com/changesets/changesets/blob/master/docs/adding-a-changeset.md) or give me access to your fork so I can do it for you?

packages/core/src/sensors/pointer/AbstractPointerSensor.ts
```diff
@@ -172,6 +173,12 @@ export class AbstractPointerSensor implements SensorInstance {
       }
 
       if (isDistanceConstraint(activationConstraint)) {
+        if (
+          activationConstraint.tolerance &&
```
clauderic, line 177:
```suggestion
          activationConstraint.tolerance != null &&
```

PR #408, opened 2021-08-10
fix: compute collision against DragOverlay child bounding rect

### LINKED ISSUE #401: Allow collision with drag overlay node rather than activeNode
In our application it's a frequent occurrence that things that can be dragged/sorted together are of different sizes. These sizes can differ radically and unpredictably and therefore make the dragging experience with a "presentational component" of the underlying node pretty bad. As an example, imagine dragging a giant rectangle from one container grid into a different container grid of many tiny squares. The large rectangle covers many of the squares during the drag, making it hard, from both the user's perspective and for the collision detection, to precisely drop where you want.

For this reason, and to provide additional contextual information about the thing being dragged, we've opted to go with a statically sized drag overlay that has no relation to the underlying node being dragging. It's small in size, contains an icon indicating what's being dragged, and the number of items that will ultimately be affected by the drag.

Our issue is that `DragOverlay` is sized based on the underlying node that started the drag. We can fix this stylistically by passing a style prop, but the `ViewRect` that is passed into collision detection is still from the underlying `activeNode`. This makes collision computation inaccurate in relation to what the user actually sees (our small, statically sized child of `DragOverlay`). 

I can see a few paths forward for making this experience better, but would love to get @clauderic's input here as, currently, `DragOverlay` seems fundamentally tied to the `activeNode` in many ways. Ultimately, the high-level goal is to allow collision to occur with drag overlay rather than the node the started the drag.

wmain:
Fixes #401 

This PR corrects what is, I believe, an improper assumption made in `DnDContext`. Defending this decision is most expeditiously done by providing an example of bad behavior on the master branch:

collision

In this gif the black box with the text "1 item" is the first child of the `DragOverlay`, the blue box is the drag overlay itself, and the red box is the final computed collision rectangle. The drag overlay has the `snapToCenter` modifier applied, and has standard `display: flex` + centering styles applied to center the overlay child within the `DragOverlay`. The blue/red rectangle backgrounds exist purely for debugging/demonstration purposes and would not be visible to the end user. In practice, the user would visualize their cursor and the "1 item" element as the thing driving collision, but the current approach improperly (I believe) ignores the positioning of the overlay child and only adopts its width/height. If we remove the calculation of `rect` and just use `draggingNodeRect` the expected behavior occurs and the collision rectangle is positioned appropriately. 

It's not immediately obvious to me why this logic existed in the first place, so I wholly expect that there's a case I might be missing/am not thinking of, but ultimately this _feels_ like the correct behavior and the behavior I would expect as a developer making use of the library. I ultimately want collision to be measured against the bounding rect of the first child of `DragOverlay`. If I want that to be equivalent to the active node then all I have to do is not modify the position of the first child of the overlay and its bounding box will be the same.

As an aside, working through this has made me think that there might be long term value in providing some configuration option on `DragOverlay`(or a new component like `PointerOverlay`) that foregoes the relationship with the active node and instead just aligns with the cursor position. It would make this case a lot simpler to implement and wouldn't require the use of the `snapToCenter` modifier.

### DIFF
--- a/.changeset/dragoverlay-collision-rect.md
+++ b/.changeset/dragoverlay-collision-rect.md
@@ -1,0 +1,5 @@
+---
+"@dnd-kit/core": minor
+---
+
+The collision rect is now completely based on the position of the `DragOverlay` when it is used. Previously, only the `width` and `height` properties of the `DragOverlay` were used for the collision rect, while the `top`, `left`, `bottom` and `right` properties were derived from the active node rect. This new approach is more aligned with developers would expect, but could cause issues for consumers that were relying on the previous (incorrect) behavior.
--- a/packages/core/src/components/DndContext/DndContext.tsx
+++ b/packages/core/src/components/DndContext/DndContext.tsx
@@ -199,7 +199,6 @@ export const DndContext = memo(function DndContext({
   const activeNodeClientRect = useClientRect(activeNode);
   const initialActiveNodeRectRef = useRef<ViewRect | null>(null);
   const initialActiveNodeRect = initialActiveNodeRectRef.current;
-  const nodeRectDelta = getRectDelta(activeNodeRect, initialActiveNodeRect);
   const sensorContext = useRef<SensorContext>({
     active: null,
     activeNode,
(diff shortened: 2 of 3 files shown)


wmain:
Unclear presently why this makes the E2E test flaky (all cases pass locally).

clauderic:
The e2e tests are sometimes flaky but they seem to be consistently failing here. I vaguely remember attempting this change a while back and running into similar issues

wmain:
Any chance you remember what you investigated there? Sort of unclear at first glance why this would make the results flaky between environments.

clauderic:
Okay I found the problem after a bit of digging.

When dragging without a drag overlay, we need to compute a delta between the draggable node's initial position and new position if it is reparented.

This logic isn't relevant when using the DragOverlay. It was working before because we were only using the width and height of the drag overlay in the collision rect.

> As an aside, working through this has made me think that there might be long term value in providing some configuration option on DragOverlay(or a new component like PointerOverlay) that foregoes the relationship with the active node and instead just aligns with the cursor position. It would make this case a lot simpler to implement and wouldn't require the use of the snapToCenter modifier.

That's an interesting idea. I would assume most consumers would still want most of the features of `DragOverlay` (drop animation for example), so I'd be a bit hesitant to create a brand new concept.

I'd lean towards extending `DragOverlay` to support this type of use-case.

wmain:
@clauderic The additional commit actually breaks the functionality achieved without it for my case. I'm looking into why.

wmain:
Adding back the `willComputeLayouts` flag to the `useClientRect` call for `overlayNodeRect` restores things to good working order. Will open a PR for discussion there as it's not obvious to me why it was removed.

PR #426, opened 2021-08-18
fix: pass willRecomputeLayouts to useClientRect call for overlayNodeRect

wmain:
Thanks for the additional investigation done for #408 @clauderic. It looks like removing `willRecomputeLayouts` from the `useClientRect` call of `overlayNodeRect` rebreaks our case, however. Reading over the methodology/cause of the changes you made, I'm not sure why this argument was removed, but adding it back in seems to fix things. It seems to me that recomputing the layout rect of the measurable child is required as the overlay moves.

### DIFF
--- a/packages/core/src/components/DndContext/DndContext.tsx
+++ b/packages/core/src/components/DndContext/DndContext.tsx
@@ -228,7 +228,8 @@ export const DndContext = memo(function DndContext({
 
   const [overlayNodeRef, setOverlayNodeRef] = useNodeRef();
   const overlayNodeRect = useClientRect(
-    activeId ? getMeasurableNode(overlayNodeRef.current) : null
+    activeId ? getMeasurableNode(overlayNodeRef.current) : null,
+    willRecomputeLayouts
   );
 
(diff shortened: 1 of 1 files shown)


wmain:
Nuts, E2E tests are broken in CI and working locally again. I assume you've got some insight about what's going on here.

clauderic:
> Nuts, E2E tests are broken in CI and working locally again. I assume you've got some insight about what's going on here.

Yeah, go try this story which illustrates the problem http://localhost:6006/iframe.html?id=presets-sortable-multiple-containers--many-items&args=&viewMode=story

When we use the DragOverlay as the source of the collision rect, everything works fine when you measure it for the first time, before it's moved using `transform`.

Subsequent measurements are affected by the `transform` that is applied on the DragOverlay, such that the `top`, `left`, `right` and `bottom` properties that we get back already include the `transform` that we're applied, which represents the net movement of the active sensor.

To fix this, I think the easiest way would be to subtract the current transform on subsequent measurements of the DragOverlay rect.

clauderic:
It's a rather involved fix so I'll open a separate PR

PR #427, opened 2021-08-18
Fix issues with DragOverlay measuring

clauderic:
This PR follows-up on #408 and #426

It fixes a number of issues with the measuring strategy for the DragOverlay. The main change is that we need to subtract the transform applied on the DragOverlay node when measuring it, otherwise the returned measurement includes the transforms, which means that the next time we compute the collision rect (by adding the transform to the drag overlay position), the newly calculated collision rect has double transforms, which is incorrect.

@wmain can you test this PR against your branch and confirm that it fixes the issues you were seeing with the original solution?


packages/sortable/src/components/SortableContext.tsx
```diff
@@ -47,14 +47,14 @@ export function SortableContext({
...
-    overlayNode,
+    dragOverlay,
     droppableRects,
     over,
     recomputeLayouts,
     willRecomputeLayouts,
   } = useDndContext();
   const containerId = useUniqueId(ID_PREFIX, id);
-  const useDragOverlay = Boolean(overlayNode.rect !== null);
+  const useDragOverlay = Boolean(dragOverlay.rect !== null);
```
wmain, line 57: This throws an error on my branch as `dragOverlay` is null.

clauderic: Yeah this happens sometimes with breaking TypeScript changes across packages, just re-run `yarn start` and it'll fix itself (eventually). Haven't had time to prioritize upgrading development tooling but it's on the roadmap.

clauderic:
I converted this PR to draft because I still need to make a couple tweaks, but it should be good enough for you to give it a try

packages/sortable/src/components/SortableContext.tsx
wmain: Yep, kicking it a few times did the trick.

wmain:
Looks good on my branch!

clauderic:
Good to hear. Out of curiosity, are you positioning the drag overlay child with `transform` or `margin` / `padding` / `top` / `left`? 

I'm a bit worried about the amount of complexity it adds to try and support measuring the drag overlay child using [`getBoundingClientRect()`](https://github.com/clauderic/dnd-kit/blob/master/packages/core/src/utilities/rect/getRect.ts#L51-L90) (because it is transformed) and am debating shifting to the `getViewRect(https://github.com/clauderic/dnd-kit/blob/master/packages/core/src/utilities/rect/getRect.ts#L92-L110)` method instead, which ignores transforms.

wmain:
> Good to hear. Out of curiosity, are you positioning the drag overlay child with `transform` or `margin` / `padding` / `top` / `left`?
> 
> I'm a bit worried about the amount of complexity it adds to try and support measuring the drag overlay child using [`getBoundingClientRect()`](https://github.com/clauderic/dnd-kit/blob/master/packages/core/src/utilities/rect/getRect.ts#L51-L90) (because it is transformed) and am debating shifting to the `getViewRect(https://github.com/clauderic/dnd-kit/blob/master/packages/core/src/utilities/rect/getRect.ts#L92-L110)` method instead, which ignores transforms.

We add `display: flex`, `alignItems: center`, and `justifyContent: center` to the `DragOverlay` styles.

clauderic:
That counts towards `offsetTop` / `offsetLeft` so would still be supported (https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetTop)

wmain:
For _my_ curiosity's sake, which step of computation are you considering shifting? Shouldn't the final rect include transformations and therefore make `getBoundingClientRect` appropriate?

clauderic:
There are a lot of timing-related concerns with regards to measuring elements that have transforms applied. It requires elaborate orchestration to make sure you measure an element at the appropriate point in time, either before or after transforms are updated.

You can see issues with the current strategy if you navigate to the Multiple containers story and sort with the keyboard. The timing of when the DragOverlay node is measured conflicts with the collision detection logic (there's a race condition between the two).


https://user-images.githubusercontent.com/1416436/129971595-fabfa2ea-530b-4c8e-84ad-9f18e5373271.mp4


These issues aren't impossible to solve, but I'd rather keep the code simple and reliable for now and improve on it later.


https://user-images.githubusercontent.com/1416436/129971617-9ea15449-4a1d-441d-a6d2-20e868f1d162.mp4

Issue #429, opened 2021-08-20
Issues with `overflow:auto` on `<html>` element


chestozo:
Hello guys
Thank you for the amazing lib!

We are integrating it on our project and we've found an issue  
with `overflow: auto` that we have set on `<html/>` element:
in this case all works just fine until user scrolls the page.

Here is a video of the issue.

I am running this story from your storybook:
`Droppable / useDroppable / Basic Setup`
http://localhost:6006/?path=/story/core-droppable-usedroppable--basic-setup

I've added `overflow: auto` on `<html/>` element:

https://user-images.githubusercontent.com/498934/130198358-69e8638a-4e1e-4cda-aa04-0ff095371109.mov

Is it something that should be fixed on app level somehow?
Or this is something that can be fixed on lib level?
Thank you

PR #430, opened 2021-08-20
Fix duplicate scroll ancestor detection

clauderic:
Fixes https://github.com/clauderic/dnd-kit/issues/429

In some scenarios, an element could be added twice to the list of scrollable ancestors, resulting in invalid offsets.


chestozo:
what a speed man! 💪

Issue #508, opened 2021-11-17
Support use in portal / modals and popout windows?


larissa-n:
Thanks for this great library! It's hard to find a drag-and-drop library that supports both flex-wrap (horizontal + vertical) and mobile out of the box. Fantastic job!

Any chance you might be supporting use in portals, like modals and popout windows (i.e. outside the DOM)? Currently this causes a crash. 
```
element.getBoundingClientRect is not a function
at /dnd-kit/core/dist/core.esm.js:768
```
Demo: https://codesandbox.io/s/dnd-kit-react-popout-portal-window-forked-dbfcj?file=/pages/index.js

clauderic:
I did a bit of digging, this should in theory be supported, but isn't working because of a bug in the code in how we check if an element is an instance of `Window` here: https://github.com/clauderic/dnd-kit/blob/master/packages/core/src/utilities/rect/getRect.ts#L54

<img width="1266" alt="Screen Shot 2021-11-17 at 3 15 16 PM" src="https://user-images.githubusercontent.com/1416436/142275571-ee004b7e-ae57-46f0-99fe-ecc3a41ee925.png">

PR #509, opened 2021-11-18
Update helpers to support multi-window use cases

clauderic:
Fixes https://github.com/clauderic/dnd-kit/issues/508


Nauss:
Hi @clauderic, I'm using dndkit in an Electron app and I must warn you that these modifications are breaking for me.

In Electron, the "window" object is actually called `global`. So in this context, `global instanceof Window === true` but  `Object.prototype.toString.call(global) === '[object global]'` so the `isWindow` function returns false and I get a crash on the next call to `global.getBoundingClientRect`.

I'm not sure how this could be solved and still fixing #508...

clauderic:
@Nauss want to try opening a PR that updates `isWindow` to also check for `[object global]`?

Nauss:
Ok, I'll have a look tomorrow

Nauss:
The PR is ready: #532.
I'm still not sure if the solution is universal but it works for my use case.

Issue #483, opened 2021-10-23
Cannot be dragged into the DropArea when the DragItem is at the bottom of the scroll container


zhang0ZGC:
I build a scroll container with many items and a drop area with fixed height。But if i drag the item at the bottom of the scroll container into drop area，the dragOver event will not be triggered, and the `over` attribute of the dragEnd event is also null.

I have no idea how to fix this situation😥

https://user-images.githubusercontent.com/10346399/138538354-06d22836-1039-4c1c-81ca-faebaa7d353f.mp4

Here is a reproduction case：

https://codesandbox.io/s/drag-item-in-scroller-e36j8

HelKyle:
Improved

```ts
import type {
  CollisionDetection,
  LayoutRect,
  UniqueIdentifier,
  ViewRect
} from '@dnd-kit/core'

function getIntersectionRatio(entry: LayoutRect, target: ViewRect): number {
  const top = Math.max(target.top, entry.offsetTop)
  const left = Math.max(target.left, entry.offsetLeft)
  const right = Math.min(
    target.left + target.width,
    entry.offsetLeft + entry.width
  )
  const bottom = Math.min(
    target.top + target.height,
    entry.offsetTop + entry.height
  )
  const width = Math.abs(right - left)
  const height = Math.abs(bottom - top)

  if (left < right && top < bottom) {
    const targetArea = target.width * target.height
    const entryArea = entry.width * entry.height
    const intersectionArea = width * height
    const intersectionRatio =
      intersectionArea / (targetArea + entryArea - intersectionArea)

    const ratio = Number(intersectionRatio.toFixed(4))
    return ratio
  }

  return 0
}

export const rectIntersection: CollisionDetection = args => {
  const { active, droppableContainers, collisionRect } = args
  let maxIntersectionRatio = 0
  let maxIntersectingDroppableContainer: UniqueIdentifier | null = null

  const { translated } = active.rect.current

  if (!translated) return maxIntersectingDroppableContainer
  for (let i = 0; i < droppableContainers.length; i += 1) {
    const droppableContainer = droppableContainers[i]
    const {
      rect: { current: rect }
    } = droppableContainer

    if (rect) {
      // should use collisionRect instead of translated if same container
      const computedRect =
        droppableContainer.data?.current?.sortable?.containerId ===
        active.data?.current?.sortable?.containerId
          ? collisionRect
          : translated
      const intersectionRatio = getIntersectionRatio(rect, computedRect)

[...]

Issue #464, opened 2021-09-30
Transformed SortableContext yields an offset drag overlay


ranbena:
When offsetting a SortableContext with css transform (for instance, when an open sidebar pushes it sideways), the drag overlay gets wrongly positioned.

<img width="469" alt="Screen Shot 2021-09-30 at 9 31 28" src="https://user-images.githubusercontent.com/486954/135399585-4b863509-b633-4668-87cd-1780e820ab1a.png">

Sandbox demo https://codesandbox.io/s/dnd-kit-offset-b2744

_Originally posted by @ranbena in https://github.com/clauderic/dnd-kit/issues/205#issuecomment-830186486_

ranbena:
Here's a demonstration:

https://user-images.githubusercontent.com/486954/116719759-1da62f00-a9e4-11eb-8e9b-08d079c2e643.mov

ranbena:
https://github.com/clauderic/dnd-kit/pull/379 allows consumers to configure how to measure the draggable node, so I gave it a shot:

```tsx
const measuringConfig = {
  draggable: {
    measure: (element) => {
      return {
        ...getBoundingClientRect(element),
        left: getViewportLayoutRect(element).left
      };
    }
  }
};
```

The result is better - there is no offset during drag 🎉
BUT the drop animation sends the overlay to an offset position.
Sandbox code: https://codesandbox.io/s/dnd-kit-offset-with-measuringconfig-6ujp6?file=/src/App.js

https://user-images.githubusercontent.com/486954/135401065-73783a0d-afcf-4202-87a2-6048994c0114.mov

ranbena:
@clauderic in my original post I provided a reproducible sandbox demo.
I now upgraded the dependencies in the demo to `@next` but the issue doesn't seem to be fixed - actually even more offset.
Am I missing something?

-----------------------

Current: https://b2744.csb.app/
Upgraded: https://fzcdy.csb.app/

clauderic:
Hey @ranbena,

The problem with your sandbox is that you're also rendering the `DragOverlay` component in the same transformed container, so it too is offset by `135px` when we measure it.

It's generally recommended that you try to render the `DragOverlay` in a portal higher up in the tree.

Here's a forked sandbox where that is the only change I made, everything seems to be working: https://codesandbox.io/s/dnd-kit-offset-forked-hbtwb?file=/src/App.js

ranbena:
Understood. I'll try that 👍

[...]

Issue #325, opened 2021-06-08
Drag Overlay and scrolling container not working after scroll


damien-monni:
I don't know if it is related to #43, so I am opening a new issue.

I am having an issue with a simple drag and drop that is using a scrolling container. I tried to replicate the issue in the most simple way I could find here: https://codesandbox.io/s/dnd-kit-scroll-issue-iwxrm?file=/src/App.js

When the container is not scrolled, drag and drop works perfectly. But as soon as you scroll the container and try to drag and drop an item, it starts behaving weirdly. Sometime it even seems to randomly stops working as you can see in the following video:

https://user-images.githubusercontent.com/6009968/121145555-605af280-c83f-11eb-8d0c-d584f4363e79.mp4

Any hint or workaround for this issue would be fantastic!

Thanks

clauderic:
This is indeed a duplicate of #43, though thanks for the detailed issue report and replication sandbox

Issue #263, opened 2021-05-10
Droppables will randomly interact when dragging over empty space on the bottom of scrolled container v2


ohayojp:
Regarding this closed issue - https://github.com/clauderic/dnd-kit/issues/239
I've created new code-sandbox without unnecessary elements - this is now just two draggables on the left and bunch of droppables inside table cells. 
https://codesandbox.io/s/zen-turing-6hizq

On full screen example - https://6hizq.csb.app/ - you can see that there is no other scrolls, the only one is needed for table-wrapper. This is of course simplified form of much more complex layout, so scroll cannot be removed from this element.
I would appreciate it if you could look into it.

https://user-images.githubusercontent.com/6364234/117653354-93ed2300-b194-11eb-9369-33a1451a2f1e.mov

PR #518, opened 2021-11-22
Refactor measuring and collision detection

clauderic:
This PR reduces the number of concepts related to measuring from `ViewRect`, `LayoutRect` to just a single concept of `ClientRect`.

Previously, all collision detection algorithms were relative to the top and left points of the document. While this approach worked in most situations, it broke down in a number of different use-cases: #43, #73, #98, #263, #325, #483, etc.

This new approach changes the frame of comparison to be relative to the viewport. This is a major breaking change, and will need to be released under a new major version bump.

This PR also introduces a changed approach to measuring the `ClientRect` of draggable and droppable elements. Previously, the `offsetTop` and `offsetLeft` properties of an element and all of its positioned ancestors were used to compute the rect that was used for collision detection. This approach had a number of limitations however, and failed when the draggable or droppable elements were positioned within transformed elements (since `offsetTop` and `offsetLeft` properties do not take into account transforms): #464 

By default, @dnd-kit now ignores only the `transforms` applied to the draggable / droppable node itself, but considers all the transforms applied to its ancestors. This should provide the right balance of flexibility for most consumers. 

Transforms applied to the droppable and draggable nodes are ignored by default, because the recommended approach for moving items on the screen is to use the `transform` property, which can interfere with the calculation of collisions. 

Consumers can choose an alternate approach that does consider transforms for specific use-cases if needed by configuring the `measuring` prop of `<DndContext>`. Refer to the `<Switch>` example.

### DIFF
--- a/.changeset/refactor-collision-detection.md
+++ b/.changeset/refactor-collision-detection.md
@@ -1,0 +1,28 @@
+---
+"@dnd-kit/core": major
+"@dnd-kit/sortable": major
+"@dnd-kit/utilities": minor
+"@dnd-kit/modifiers": minor
+---
+
+Major internal refactor of measuring and collision detection. 
+
+### Summary of changes
+Previously, all collision detection algorithms were relative to the top and left points of the document. While this approach worked in most situations, it broke down in a number of different use-cases, such as fixed position droppable containers and trying to drag between containers that had different scroll positions.
+
+This new approach changes the frame of comparison to be relative to the viewport. This is a major breaking change, and will need to be released under a new major version bump.
+
+### Breaking changes:
+
+- By default, `@dnd-kit` now ignores only the transforms applied to the draggable / droppable node itself, but considers all the transforms applied to its ancestors. This should provide the right balance of flexibility for most consumers.
+  - Transforms applied to the droppable and draggable nodes are ignored by default, because the recommended approach for moving items on the screen is to use the transform property, which can interfere with the calculation of collisions.
+  - Consumers can choose an alternate approach that does consider transforms for specific use-cases if needed by configuring the measuring prop of <DndContext>. Refer to the <Switch> example.
+- Reduced the number of concepts related to measuring from `ViewRect`, `LayoutRect` to just a single concept of `ClientRect`.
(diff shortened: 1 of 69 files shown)


baptisteArno:
I'm eagerly waiting for this update 💪

Let me know if I can help in any way @clauderic

CodexpathCommunity:
how can I use this solution in my app when I'm having container detection with long scrolling container

prallhsu:
> This PR reduces the number of concepts related to measuring from `ViewRect`, `LayoutRect` to just a single concept of `ClientRect`.
> 
> Previously, all collision detection algorithms were relative to the top and left points of the document. While this approach worked in most situations, it broke down in a number of different use-cases: #43, #73, #98, #263, #325, #483, etc.
> 
> This new approach changes the frame of comparison to be relative to the viewport. This is a major breaking change, and will need to be released under a new major version bump.
> 
> This PR also introduces a changed approach to measuring the `ClientRect` of draggable and droppable elements. Previously, the `offsetTop` and `offsetLeft` properties of an element and all of its positioned ancestors were used to compute the rect that was used for collision detection. This approach had a number of limitations however, and failed when the draggable or droppable elements were positioned within transformed elements (since `offsetTop` and `offsetLeft` properties do not take into account transforms): #464
> 
> By default, @dnd-kit now ignores only the `transforms` applied to the draggable / droppable node itself, but considers all the transforms applied to its ancestors. This should provide the right balance of flexibility for most consumers.
> 
> Transforms applied to the droppable and draggable nodes are ignored by default, because the recommended approach for moving items on the screen is to use the `transform` property, which can interfere with the calculation of collisions.
> 
> Consumers can choose an alternate approach that does consider transforms for specific use-cases if needed by configuring the `measuring` prop of `<DndContext>`. Refer to the `<Switch>` example.

How can i config DndContext to apply  transforms property on my droppable node itself，i'm confusing with the `<Switch>` example，and i had tried to config measuring property，but it's not working. @clauderic 
`measuring={{
        droppable: {
          measure: (element: HTMLElement) => element.getBoundingClientRect(),
        },
      }}`

prallhsu:
Sry，it works fine，but my droppable area has a transform animation when i begin to drag. So when i add `frequency` props in measuring config， everything works fine!

PR #532, opened 2021-11-30
fix: correctly check whether an element is a window in electron context

Nauss:
Fixes https://github.com/clauderic/dnd-kit/issues/540

This PR just adds a check in the isWindow function to make sure the window object is correctly checked in an Electron context.


clauderic:
Thanks for your contribution! ❤️

PR #541, opened 2021-12-12
fix: Make isWindow work for electron

### LINKED ISSUE #540: Crashes in Electron
When this module is used in Electron, it will crash when it gets to this part: [/packages/core/src/utilities/rect/getRect.ts#L80](https://github.com/clauderic/dnd-kit/blob/master/packages/core/src/utilities/rect/getRect.ts#L80)

Because the `isWindow` check will return false for Electron's `window` object, and will try to call `getBoundingClientRect` on it which will raise an error.

knubie:
Updates the `isWindow` check to also pass for Electron window objects.

Closes #540

### DIFF
--- a/packages/utilities/src/type-guards/isWindow.ts
+++ b/packages/utilities/src/type-guards/isWindow.ts
@@ -1,3 +1,4 @@
 export function isWindow(element: Object): element is typeof window {
-  return Object.prototype.toString.call(element) === '[object Window]';
+  let str = Object.prototype.toString.call(element)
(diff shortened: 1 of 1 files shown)


clauderic:
Hey @knubie, thanks for opening a PR. There's actually already a PR open that implements the same fix here https://github.com/clauderic/dnd-kit/pull/532. It was opened before yours, so I'll be merging that one.

knubie:
Great, sorry for the noise!

PR #556, opened 2022-01-04
Add pointer coordinates to collision detection

avgrad:
- added the current pointer coordinates to the collision detection feature
- implemented a `pointerWithin` collision algorithm (incl. Storybook)

With pointer coordinates, a droppable beneath the pointer can be detected for collision instead of relying only on the rectangles. In some cases this can be useful for more precice detection.

Consider use cases like a file explorer. The pointer is how for example Google Drive detects if a dragged item is above a droppable.

### DIFF
--- a/.changeset/pointer-within-algorithm.md
+++ b/.changeset/pointer-within-algorithm.md
@@ -1,0 +1,6 @@
+---
+'@dnd-kit/core': minor
+---
+
+- Added pointer coordinates to collision detection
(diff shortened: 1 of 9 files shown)


clauderic (approved):
Thanks for your contribution @avgrad, this is an excellent PR and a great addition! I rebased your PR against master since a lot of changes have just been introduced with #518.

PR #642, opened 2022-03-07
fix SortableContext performance

vosatom:
Is there any reason to recompute `sortedRects` on every update?

At the moment, `sortedRects` causes context change. This will result in updating all components using `useSortable` hook on every `pointermove` event.

### DIFF
--- a/.changeset/fix-sortablecontext-performance.md
+++ b/.changeset/fix-sortablecontext-performance.md
@@ -1,0 +1,5 @@
+---
+"@dnd-kit/sortable": patch
+---
+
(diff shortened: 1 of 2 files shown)


vosatom:
To test the difference, open the storybook for example at http://localhost:6006/?path=/story/presets-sortable-vertical--basic-setup and use React Devtools Profiler. Try dragging the "1" box, for example. Then compare with `master` branch.

**Before:** 
before

**After:**
after

clauderic (approved):
Good catch! Thanks for your contribution 🙏

PR #646, opened 2022-03-09
expose `DraggableAttributes`

### LINKED ISSUE #645: would be nice to export the attributes type so they could be easily typed in props
Sometimes we want to pass listeners and attributes into components so they can be used in drag handles. ([see flexibility](https://docs.dndkit.com/api-documentation/draggable/usedraggable#flexibility))

https://github.com/clauderic/dnd-kit/blob/6f307806c09d4a1147073bf04f7cd15b7de721fc/packages/core/src/hooks/useDraggable.ts#L18-L22

Listeners has an exposed type `DraggableSyntheticListeners` but attributes does not.
It would be nice if attributes was exposed as a public type.

lukesmurray:
Fixes #645 by exposing draggable attributes.

### DIFF
--- a/packages/core/src/hooks/useDraggable.ts
+++ b/packages/core/src/hooks/useDraggable.ts
@@ -22,6 +22,14 @@ export interface UseDraggableArguments {
   };
 }
 
+export type DraggableAttributes = {
+  role: string;
+  tabIndex: number;
+  'aria-pressed': boolean | undefined;
+  'aria-roledescription': string;
+  'aria-describedby': string;
+};
+
 export type DraggableSyntheticListeners = SyntheticListenerMap | undefined;
 
 const NullContext = createContext<any>(null);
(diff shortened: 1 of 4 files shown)


packages/core/src/hooks/useDraggable.ts
```diff
@@ -15,13 +15,15 @@ export interface UseDraggableArguments {
...
-  };
+  attributes?: DraggableAttributes;
 }
 
+export type DraggableAttributes = {
+  role?: string;
+  roleDescription?: string;
+  tabIndex?: number;
+};
+
```
clauderic, line 26: Are you interested in the argument type or the type that is returned by `useDraggable`? They are different, refer to what is returned by `memoizedAttributes` below

lukesmurray: thanks for checking I overwrote the commit with the right change.

clauderic (approved):
Thanks for your contribution 🙏

PR #676, opened 2022-03-24
allow <DragOverlay /> consumers to provide custom drop animation

### LINKED ISSUE #129: DragOverlay drop animation final destination
When using the `<DragOverlay />` component, I'd like to be able to control the `x,y` position that the overlay returns to _while_ the finishing dropAnimation is in progress.

Right now, it seems to return back to it's original position.

For some drop zones, I'd prefer it stay in the same position and just change size (e.g. trashing an item transitions its scale to 0 around the transformOrigin).

Any way to accomplish this? I tried using my own `modifiers`, but it doesn't seem to have an effect after the drop happens.

paynecodes:
Allow `<DragOverlay />` consumers to provide their own `useDropAnimation` hook. This is a fairly risky API as I've never encountered passing hooks as props. The alternative is to really think hard about the `useDropAnimation` API and offer consumers different places to "hook into" the hook. ;)

***

This is just a stab at #129 . It does work, but I'm open to suggestions.

Fixes #129

### DIFF
--- a/packages/core/src/components/DragOverlay/DragOverlay.tsx
+++ b/packages/core/src/components/DragOverlay/DragOverlay.tsx
@@ -18,6 +18,7 @@ export interface Props {
   children?: React.ReactNode;
   className?: string;
   dropAnimation?: DropAnimation | null | undefined;
+  useCustomDropAnimation?: typeof useDropAnimation | undefined;
   style?: React.CSSProperties;
   transition?: string | TransitionGetter;
   modifiers?: Modifiers;
@@ -43,6 +44,7 @@ export const DragOverlay = React.memo(
     adjustScale = false,
     children,
     dropAnimation = defaultDropAnimation,
+    useCustomDropAnimation = useDropAnimation,
     style: styleProp,
     transition = defaultTransition,
     modifiers,
@@ -147,7 +149,7 @@ export const DragOverlay = React.memo(
     const {children: finalChildren, transform: _, ...otherAttributes} =
       derivedAttributes ?? {};
     const prevActiveId = useRef(active?.id ?? null);
-    const dropAnimationComplete = useDropAnimation({
+    const dropAnimationComplete = useCustomDropAnimation({
       animate: Boolean(dropAnimation && prevActiveId.current && !active),
       adjustScale,
       activeId: prevActiveId.current,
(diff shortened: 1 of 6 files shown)


paynecodes:
@clauderic Any chance I could get some feedback on the approach here?

clauderic:
Hey @paynecodes, thanks for opening a PR.

Letting consumers provide a custom drop animation is definitely something that's planned, but before that can happen the `<DragOverlay>` component needs a bit of work. I have a number of different refactors I'm planning for the `<DragOverlay>` component, and I fear that opening the possibility for consumers to pass in a custom drop animation hook wouldn't be something that would work once `<DragOverlay>` is refactored, so I'd prefer not to introduce that complexity at this stage.

I'd recommend forking `<DragOverlay>` for now if you absolutely need this functionality.

paynecodes:
Ah, I see. Forking DragOverlay is a viable option for now. Thank you for the feedback.

paynecodes:
@clauderic How do you feel about exposing all of the currently private imports so that implementing our own `<DragOverlay />` component is feasible? 

e.g.

```
InternalContext
getRelativeTransformOrigin
applyModifiers
ActiveDraggableContext
etc...
```

Perhaps I'm missing something simple. Please advise.

clauderic:
Could you help me better understand what you're trying to achieve? What kind of "custom" drop animation are you trying to implementing? What's missing from the base drop animation?

paynecodes:
@clauderic In my application, I need to drop a draggable to specific coordinates inside the droppable container. A reasonably good example is the [story I added in my PR](https://github.com/paynecodes/dnd-kit/blob/16529812e2f7474373444397a5a25619bfdd343c/stories/1%20-%20Core/Draggable/2-DragOverlay.story.tsx#L148-L275) to drop the draggable in the center of the screen.

I don't have the full specification of everything I may want to do with this animation because I'm early in the prototyping phase. Because of this, I thought it was useful to be able to take complete control over the animation. Allowing a custom hook to be passed as props probably breaks an unwritten (as far as I can tell) rule of React Hooks somewhere. In that case, I'd love to hear some further ideas around offering an API which allows `<DragOverlay />` consumers to completely control the drop animation.

clauderic:
Refer to https://github.com/clauderic/dnd-kit/pull/733, I think this should provide plenty of flexibility to cover most use cases.

PR #690, opened 2022-04-04
Preserve active data in case of active unmounting while dragging

creatyvtype:
When using a virtualized list, the dragged item may become unrendered, losing the draggable node as referenced in the affected lines changed.

This PR preserves the node data so that even if unmounted we have access to it.

Why is this relevant?

I found that when writing a custom collision detector, in order to do any filtering against droppable containers based on the active item, I discovered the data disappeared, causing a mixture of collisions with containers I do not intend this item to collide with.

### DIFF
--- a/packages/core/src/components/DndContext/DndContext.tsx
+++ b/packages/core/src/components/DndContext/DndContext.tsx
@@ -160,11 +160,25 @@ export const DndContext = memo(function DndContext({
     event: null,
   }));
   const [isDragging, setIsDragging] = useState(false);
+  const [nodeData, setNodeData] = useState<DataRef | null>(null);
   const {
     draggable: {active: activeId, nodes: draggableNodes, translate},
     droppable: {containers: droppableContainers},
   } = state;
-  const node = activeId ? draggableNodes[activeId] : null;
+  const node = useMemo(() => (activeId ? draggableNodes[activeId] : null), [
+    activeId,
+    draggableNodes,
+  ]);
+  useEffect(() => {
+    // It's possible for the active node to unmount while dragging
+    // This will preserve the data
+    if (activeId && node?.data) {
+      setNodeData(node.data);
+    } else if (!activeId && !node) {
+      setNodeData(null);
+    }
+  }, [activeId, node]);
+
   const activeRects = useRef<Active['rect']['current']>({
     initial: null,
     translated: null,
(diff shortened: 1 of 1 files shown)


packages/core/src/components/DndContext/DndContext.tsx
```diff
@@ -160,11 +160,25 @@ export const DndContext = memo(function DndContext({
...
+  ]);
+  useEffect(() => {
+    // It's possible for the active node to unmount while dragging
+    // This will preserve the data
+    if (activeId && node?.data) {
+      setNodeData(node.data);
+    } else if (!activeId && !node) {
+      setNodeData(null);
+    }
+  }, [activeId, node]);
```
clauderic, line 180: Do you need to use state to store the reference to the node's data? Could this be achieved using `useMemo` instead? I'd like to avoid introducing additional re-renders of `<DndContext>` if possible, not just for performance reasons, but also because it can lead to `nodeData` being `null` or stale for the first render when dragging begins and the final render when dragging ends.

packages/core/src/components/DndContext/DndContext.tsx
```diff
@@ -160,11 +160,25 @@ export const DndContext = memo(function DndContext({
     event: null,
   }));
   const [isDragging, setIsDragging] = useState(false);
+  const [nodeData, setNodeData] = useState<DataRef | null>(null);
   const {
     draggable: {active: activeId, nodes: draggableNodes, translate},
     droppable: {containers: droppableContainers},
   } = state;
-  const node = activeId ? draggableNodes[activeId] : null;
+  const node = useMemo(() => (activeId ? draggableNodes[activeId] : null), [
```
clauderic, line 168: Is it necessary to memoize the `node`? This can technically change if the draggable node unmounts and re-mounts while dragging and lead to a stale reference.

clauderic (commented):
Hey @creatyvtype, thanks for opening a PR! Just have a few questions/comments about the implementation, but overall this is definitely something that could happen with virtualization, so it makes sense to introduce a fix to this issue.

porkopek:
For anybody that found this issue searching for a workaround, in the meantime this PR is approved (or not), you can save the active value in `onDragStart` and rely on this value instead of the actual value offered in `onDragEnd`, because in a virtual list might be null

PR #734, opened 2022-05-13
add runtime typechecks for sortable IDs and items.

### LINKED ISSUE #606: Throw a descriptive error in development when non-string IDs are passed to useSortable or SortableContext
The current behavior of the library when passed non-string IDs for sortable items can be confusing, see https://github.com/clauderic/dnd-kit/issues/604 and https://github.com/clauderic/dnd-kit/discussions/357#discussioncomment-1018851 for additional context. 

I plan to add runtime checks to the useSortable and SortableContext hook that will throw a descriptive error in development environments only when the `id` passed to `useSortable` is a non-string value, or the `items` array passed to `<SortableContext>` contains non-string values. Hopefully this will save future users of the library the debugging headaches that I went through over the past couple of days.

qrohlf:
Marking this as a draft because I still need to NPM link this into my project and manually test it. 

Resolves #606

### DIFF
--- a/packages/sortable/src/components/SortableContext.tsx
+++ b/packages/sortable/src/components/SortableContext.tsx
@@ -60,6 +60,21 @@ export function SortableContext({
       ),
     [userDefinedItems]
   );
+  // add a runtime type check for non-TS users to avoid confusion around
+  // non-string IDs. See the following links for context on why this is useful.
+  // https://github.com/clauderic/dnd-kit/issues/604
+  // https://github.com/clauderic/dnd-kit/discussions/357#discussioncomment-1018851
+  // https://github.com/clauderic/dnd-kit/issues/606
+  //
+  // the process.env guard will cause this check to be removed from the bundle
+  // at build time by most modern build tools
+  if (process.env.NODE_ENV !== 'production') {
+    if (items.every((i) => typeof id === 'string')) {
+      throw Error(
+        'SortableContext expects items to be an array of either strings, or { id: string } objects. See https://docs.dndkit.com/presets/sortable/sortable-context#items'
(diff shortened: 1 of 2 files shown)


packages/sortable/src/components/SortableContext.tsx
```diff
@@ -60,6 +60,21 @@ export function SortableContext({
...
+  // add a runtime type check for non-TS users to avoid confusion around
+  // non-string IDs. See the following links for context on why this is useful.
+  // https://github.com/clauderic/dnd-kit/issues/604
+  // https://github.com/clauderic/dnd-kit/discussions/357#discussioncomment-1018851
+  // https://github.com/clauderic/dnd-kit/issues/606
+  //
+  // the process.env guard will cause this check to be removed from the bundle
+  // at build time by most modern build tools
+  if (process.env.NODE_ENV !== 'production') {
+    if (items.every((i) => typeof id === 'string')) {
```
qrohlf, line 72: Because SortableContext is relatively high up in the render tree, I don't _think_ this will cause perf issues in development but because it's iterating over the items array, it does have the potential to.

This check could be memoized but it would need to use something besides React.useMemo in order to avoid the hooks linter complaining about hooks inside a conditional (even if that conditional will never change at runtime...)

clauderic:
Hey @qrohlf, thanks for opening a PR ❤️ 

I have a different solution in mind to this problem so I will be closing this PR for the time being https://github.com/clauderic/dnd-kit/pull/755

Feel free to re-open an issue to discuss if you feel like https://github.com/clauderic/dnd-kit/pull/755 doesn't fully address your concerns.

PR #740, opened 2022-05-17
Placeholders

5ebastianMeier:
@clauderic first of all:
Thanks for this fantastic lib!
I've invested so much time to adapt and marry `react-dnd` and `react-beautiful-dnd` before realizing that `dnd-kit` may be able to make my life a LOT easier.

## Motivation
When I first started using `dnd-kit` I didn't even notice that placeholders were not supported that prominently, because in the app I maintain most drop zones only show a full overlay over the dropzone when you drag an element over them that can be dropped.

In a common sortable scenario where you only move an existing entry around placeholders are also not a problem.
But when I tried to implement a dynamic list where entries can be added and sorted in the same step via drag-n-drop I started to realize that this can be very complicated. 

Especially when you use custom appearances for one or more of the elements participating in the operation (e.g. custom drag overlay that represents an abstract type, but becomes a specialized type once you move it onto a droppable). Furthermore I felt a lack of control to independently style and customize certain behaviors.

Another thing that I noticed is that drop animations are limited in terms of animating to a drop target, because a drag mostly seems to be treated as a "move" operation. In case of a "copy" operation the placeholder can also serve as a target to animate the drop to.

## Demo

I've added two Stories that demonstrate the behaviors this PR enables.

These videos give you a glimpse of the functionality - feel free to check out the branch and try it yourself.

### Placeholder for a regular Droppable
https://user-images.githubusercontent.com/16539774/168794615-110769a7-f9ce-4d75-8dff-511f32f713e5.mov

### Dynamic Placeholder for a Sortable
https://user-images.githubusercontent.com/16539774/168794940-8d7a18f2-4c6a-4ac9-993e-b9ffe08d1fcc.mov

### Dynamic Placeholder with a custom height DragOverlay
https://user-images.githubusercontent.com/16539774/168795026-d9ac4a27-89a5-49eb-8498-5870f267fcd6.mov

### Dynamic Placeholder that tracks the custom height of the DragOverlay
https://user-images.githubusercontent.com/16539774/168795104-df702731-ec2b-4256-95b6-eb98bedd5957.mov

## Implementation
This is just my naive first try to implement the behavior I was missing. I'm sure there are edge cases that I might have missed and the API can certainly be improved as well.

I'd rather use this as a starting point for the discussion. There are some dirty hacks included that I didn't put too much effort into yet (e.g. string-replace to parse an id when I didn't have a good idea how to achieve it in a clean fashion).

### DIFF
--- a/packages/core/src/components/DndContext/DndContext.tsx
+++ b/packages/core/src/components/DndContext/DndContext.tsx
@@ -543,6 +543,8 @@ export const DndContext = memo(function DndContext({
               rect: overContainer.rect.current,
               data: overContainer.data,
               disabled: overContainer.disabled,
+              placeholderId: overContainer.placeholderDraggableId,
+              placeholderContainerId: overContainer.placeholderContainerId,
             }
           : null;
       const event: DragOverEvent = {
--- a/packages/core/src/components/DragOverlay/DragOverlay.tsx
+++ b/packages/core/src/components/DragOverlay/DragOverlay.tsx
@@ -150,7 +150,10 @@ export const DragOverlay = React.memo(
     const dropAnimationComplete = useDropAnimation({
       animate: Boolean(dropAnimation && prevActiveId.current && !active),
       adjustScale,
-      activeId: prevActiveId.current,
+      activeId:
+        over && over.placeholderId.current
+          ? over.placeholderId.current
+          : prevActiveId.current,
       draggableNodes,
       duration: dropAnimation?.duration,
       easing: dropAnimation?.easing,
--- a/packages/core/src/hooks/useDraggable.ts
+++ b/packages/core/src/hooks/useDraggable.ts
@@ -20,6 +20,8 @@ export interface UseDraggableArguments {
     roleDescription?: string;
     tabIndex?: number;
   };
+  placeholder?: boolean;
+  placeholderContainerId?: string;
 }
 
 export interface DraggableAttributes {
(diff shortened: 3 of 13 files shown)


clauderic:
Hey @5ebastianMeier, thanks for opening a PR and for providing detailed motivation and rationales for this change ❤️ 

I'm a bit reluctant to introduce this type of complexity into the built-in concepts of @dnd-kit.

There are already a lot of concepts that consumers need to understand, and I'm not sure whether this is unlocking net new functionality that can't otherwise be achieved with the existing concepts that @dnd-kit provides out of the box.

None of the existing @dnd-kit primitives are opinionated about rendering; introducing the concept of placeholders into the built-in concepts feels like it would blur the lines between the consumer being fully in control of rendering and state management.

Having that said, I totally see the value in making this type of use-case easier for consumers; I just have a different approach in mind for how to achieve that goal.

One of the things that has been on the roadmap for quite some time but that I have not had time to prioritize yet is to create one or multiple @dnd-kit packages that would export components built on top of the core concepts of @dnd-kit and that could expose concepts such as placeholders.

```ts
import {MultipleContainers} from '@dnd-kit/sortable-components';

function App() {
  const [items, setItems] = useState({A: ['A1', 'A2', 'A3'], B: ['B1', 'B2', 'B3']});

  return (
    <MultipleContainers
      items={items}
      onItemsChange={setItems}
      renderContainer={...}
      renderItem={...}
      renderPlaceholder={...}
   />
  )
}
```

I'm very appreciative of the time and effort you've put into this PR, it looks very polished. In the future I would recommend starting a discussion beforehand just to align on whether or not the functionality you have in mind fits within the goals of the library.

clauderic:

PR #754, opened 2022-05-20
Introduce `disabled` prop on SortableContext

clauderic:
The `<SortableContext>` component now optionally accepts a `disabled` prop to globally disable `useSortable` hooks rendered within it.

The `disabled` prop accepts either a boolean or an object with the following shape:

```ts
interface Disabled {
  draggable?: boolean;
  droppable?: boolean;
}
```

The `useSortable` hook has now been updated to also optionally accept the `disabled` configuration object to conditionally disable the `useDraggable` and/or `useDroppable` hooks used internally.

```ts
useSortable({
  disabled: {
    draggable: true,
    droppable: false,
  }
});
```

Like the `strategy` prop, the `disabled` prop defined on the `useSortable` hook takes precedence over the `disabled` prop defined on the parent `<SortableContext>`.

### DIFF
--- a/.changeset/sortable-disabled-prop.md
+++ b/.changeset/sortable-disabled-prop.md
@@ -1,0 +1,18 @@
+---
+'@dnd-kit/sortable': minor
+---
+
+The `<SortableContext>` component now optionally accepts a `disabled` prop to globally disable `useSortable` hooks rendered within it.
+
+The `disabled` prop accepts either a boolean or an object with the following shape:
+
+```ts
+interface Disabled {
+  draggable?: boolean;
+  droppable?: boolean;
+}
+```
+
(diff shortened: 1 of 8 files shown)


maxpaleo:
Hi ! 

We are currently implementing DND within an application and need to control which draggable elements can be dragged in specific droppable elements. 

The docs (section: [useDroppable](https://docs.dndkit.com/api-documentation/droppable/usedroppable)) show an example using the data argument with a custom "accepts" parameter. This is exactly what we need, but I am now trying to figure out how we set the restriction. 

Do we need to set the droppable element to "Disabled = true" if the droppable item doesn't accept the item? Or is there a better way? 

```
function Droppable() {
  const {setNodeRef} = useDroppable({
    id: 'droppable',
    data: {
      accepts: ['type1', 'type2'],
    },
  });

  /* ... */
}
```

PR #797, opened 2022-06-14
Fix a regression with the KeyboardSensor's scrolling logic

clauderic:
There was a regression introduced in https://github.com/clauderic/dnd-kit/commit/750d72655922363b2218d7b41e028f9dceaef013 with the `KeyboardSensor`'s scrolling logic.

When the KeyboardSensor can scroll to the new coordinates but there is a delta on the opposite axis (for example, if we can scroll on the `y` axis but there is a delta on the `x` axis) the `KeyboardSensor` would scroll to the `maxScroll` or `minScroll` coordinates instead of the new scroll coordinates.

### DIFF
--- a/packages/core/src/sensors/keyboard/KeyboardSensor.ts
+++ b/packages/core/src/sensors/keyboard/KeyboardSensor.ts
@@ -125,14 +125,8 @@ export class KeyboardSensor implements SensorInstance {
 
         for (const scrollContainer of scrollableAncestors) {
           const direction = event.code;
-          const {
-            isTop,
-            isRight,
-            isLeft,
-            isBottom,
-            maxScroll,
-            minScroll,
-          } = getScrollPosition(scrollContainer);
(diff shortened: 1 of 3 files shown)


packages/core/src/sensors/keyboard/KeyboardSensor.ts
```diff
@@ -215,20 +215,33 @@ export class KeyboardSensor implements SensorInstance {
...
+                top: -scrollDelta.y,
+                behavior: scrollBehavior,
+              });
+            }
 
             break;
           }
         }
 
+        console.log(
```
alexandcote, line 238: 🔥

clauderic: https://github.com/clauderic/dnd-kit/commit/4a5132db4e6acbe407df3d3e2fd18b466647746d

PR #896, opened 2022-09-28
Don't require allowSyntheticDefaultImports: true

apepper:
`dnd-kit` currently ships with typescript type that will fail, if you do have `allowSyntheticDefaultImports` set to `false`:

```
node_modules/@dnd-kit/core/dist/components/DndContext/DndContext.d.ts:1:8 - error TS1259: Module '"node_modules/@types/react/index"' can only be default-imported using the 'allowSyntheticDefaultImports' flag

1 import React from 'react';
         ~~~~~

  node_modules/@types/react/index.d.ts:62:1
    62 export = React;
       ~~~~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'allowSyntheticDefaultImports' flag
```

This PR changes imports of React to make it compatible with allowSyntheticDefaultImports: false.

### DIFF
--- a/packages/accessibility/src/components/HiddenText/HiddenText.tsx
+++ b/packages/accessibility/src/components/HiddenText/HiddenText.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import * as React from 'react';
 
 interface Props {
   id: string;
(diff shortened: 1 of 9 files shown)


apepper:
There seems to not be any interest in this PR => closing.

PR #906, opened 2022-10-05
more efficient new index getter

### LINKED ISSUE #905: proposed improvement to defaultNewIndexGetter
The default new index getter doesn't have to copy the array to compute the new index.

https://github.com/clauderic/dnd-kit/blob/e6d544f3f775e6a43de25d1aee67efeec0d5db58/packages/sortable/src/hooks/defaults.ts#L11-L16

The following code checks two cases

1. activeIndex moved from after the index to before the index, so the index should shift up by 1.
2. activeIndex moved from before the index to after the index, so the index should shift down by 1.

```ts
export const defaultNewIndexGetter: NewIndexGetter = ({
  id,
  items,
  activeIndex,
  overIndex,
}) => {
  const index = items.indexOf(id);

  if (activeIndex > index && overIndex < index) {
    return index + 1;
  }
  if (activeIndex < index && overIndex > index) {
    return index - 1;
  }
  return index;
};
```

Ran into this on virtualized tables with 50K rows. Copying the array for each visible row was the largest bottleneck. 
If the user can get the index in O(1) you could expose a helper method to compute this in O(1).

```ts
export function newIndexHelper(
  getIndexById: (id: UniqueIdentifier) => number
): NewIndexGetter {
  return ({ id, activeIndex, overIndex }) => {
    const index = getIndexById(id);

    if (activeIndex > index && overIndex < index) {
      return index + 1;
    }
    if (activeIndex < index && overIndex > index) {
      return index - 1;
    }
    return index;
  };
}
```

lukesmurray:
fixes #905

### DIFF
--- a/packages/sortable/src/hooks/defaults.ts
+++ b/packages/sortable/src/hooks/defaults.ts
@@ -1,7 +1,4 @@
 import {CSS} from '@dnd-kit/utilities';
-
-import {arrayMove} from '../utilities';
-
 import type {
   AnimateLayoutChanges,
   NewIndexGetter,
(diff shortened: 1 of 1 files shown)


clauderic:

PR #907, opened 2022-10-05
more efficient containers scheduled for measurement

lukesmurray:
Using a set instead of an array is about 10x faster for a list with 50k items during scroll.


omril1:
Does it actually help in this case? Care to share some benchmarks on this example?
I assume the claim does not involve recreating the set every time, right?

lukesmurray:
yup doesn't hold up at all on lower level benchmarks which is the more common use case so closing.

PR #908, opened 2022-10-05
allow users to pass a more efficient method to compute the item index

lukesmurray:
add a new optional prop to `SortableContext` to let users compute the index of the item themselves. 
This is especially useful if the user can compute the index in O(1) for large lists vs the default O(n).


clauderic:

Issue #877, opened 2022-08-31
Error buliding with TypeScript 4.8


benasher44:
Hi there! We're getting the following error building the latest version (6.0.5) of the core package:

```
Error: node_modules/@dnd-kit/core/dist/sensors/types.d.ts(56,20): error TS2344: Type 'T' does not satisfy the constraint 'Object'.
```

It looks like one of the generic T parameters is missing an `extends Object` or `extends {}`, which TS 4.8 is stricter about.

dcousens:
Relevant typescript
https://github.com/clauderic/dnd-kit/blob/3abe4f13a3db7c3c8257a7175728cf9a8e1d577c/packages/core/src/sensors/types.ts#L76-L87

See
```typescript
sensor: Sensor<T>; 
```

Cannot be satisfied when `T` is **not** refined to `extend Object`.

> (type parameter) T in type SensorDescriptor<T>
> Type 'T' does not satisfy the constraint 'Object'.
> ... This type parameter might need an `extends Object` constraint.

Ayc0:
I can propose a PR

PR #948, opened 2022-11-22
Upgrade to TS 4.8

Ayc0:
Fixes https://github.com/clauderic/dnd-kit/issues/877

- Upgrade from TS 4.6 to TS 4.8 (I could have upgraded to TS 4.9 but the rest of the ecosystem like typescript-eslint isn't ready yet for it, and also the major breaking change is from the 4.7 to the 4.8, so we should have just 1 release for the 4.8 update to smooth the upgrade)
- Fix 2 type issues:
  - one regarding the generic in `SensorDescription` (it now extends `Object`)
  - another one regarding the previously untyped dependencies array in `useLatestValue`

## Test

Those are only type changes, so it cannot break the runtime of the libraries.

Also I ran `yarn tsc -p . --noEmit`, before and after the updates, and
- before I had 9 errors were reported in 6 files,
- after the update (without the 2 fixes), I had 11 errors,
- and after the update and the fixes, I'm back again at 9 errors
  (as those are already here on `master`, I didn't fix them in this PR)

image

## Release

As this is just a minor patch, this could either be a patch release.

### DIFF
--- a/packages/core/src/sensors/types.ts
+++ b/packages/core/src/sensors/types.ts
@@ -81,7 +81,7 @@ export interface Sensor<T extends Object> {
 
 export type Sensors = Sensor<any>[];
 
-export type SensorDescriptor<T> = {
+export type SensorDescriptor<T extends Object> = {
   sensor: Sensor<T>;
   options: T;
 };
--- a/.changeset/dependencies-update-typescript.md
+++ b/.changeset/dependencies-update-typescript.md
@@ -1,0 +1,6 @@
+---
+"@dnd-kit/core": patch
+"@dnd-kit/utilities": patch
+---
+
+Upgrade to TypeScript to 4.8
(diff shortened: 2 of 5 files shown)


packages/core/src/sensors/types.ts
```diff
@@ -81,7 +81,7 @@ export interface Sensor<T extends Object> {
 
 export type Sensors = Sensor<any>[];
 
-export type SensorDescriptor<T> = {
+export type SensorDescriptor<T extends Object> = {
```
Ayc0, line 84: I use the same boundary is for `Sensor`:
`export interface Sensor<T extends Object> {`

Ayc0:
cc @clauderic

clauderic (approved):

Ayc0:
🙇‍♂️

PR #1073, opened 2023-03-21
fix(sensors): add touchcancel event to cancel dragging when dropdown …

spongeBor:
I fixed the error that dragging event stop when mobile dropdown menu appear.


clauderic:
Closing in favor of https://github.com/clauderic/dnd-kit/pull/1541

PR #1081, opened 2023-03-24
Only consider document scrollable if not prevented

leo:
At the moment, the `<html>` element will always be added to the list of scrollable ancestors, even if the element has its scrolling ability deactivated using `overflow: hidden`.

As a result, sensors like `KeyboardSensor` will mistakenly scroll `<html>` when dragging elements, causing unwanted glitches in the visual experience.

### DIFF
--- a/packages/core/src/utilities/scroll/hasOverflowStyles.ts
+++ b/packages/core/src/utilities/scroll/hasOverflowStyles.ts
@@ -1,12 +1,12 @@
 import {getWindow} from '@dnd-kit/utilities';
 
-export function isScrollable(
-  element: HTMLElement,
+export function hasOverflowStyles(
+  element: Element,
   computedStyle: CSSStyleDeclaration = getWindow(element).getComputedStyle(
     element
-  )
+  ),
+  overflowRegex: RegExp
 ): boolean {
-  const overflowRegex = /(auto|scroll|overlay)/;
   const properties = ['overflow', 'overflowX', 'overflowY'];
 
   return properties.some((property) => {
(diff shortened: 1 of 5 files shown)


packages/core/src/utilities/scroll/hasOverflowStyles.ts
```diff
@@ -1,12 +1,12 @@
 import {getWindow} from '@dnd-kit/utilities';
 
-export function isScrollable(
-  element: HTMLElement,
+export function hasOverflowStyles(
+  element: Element,
```
leo, line 4: `HTMLElement` is a subset of `Element`, so `Element` should be fine to use here (it won't break the public API).

packages/core/src/utilities/scroll/isScrollingDisabled.ts
```diff
@@ -0,0 +1,12 @@
+import {getWindow} from '@dnd-kit/utilities';
+
+import {hasOverflowStyles} from './hasOverflowStyles';
+
+export function isScrollingDisabled(
```
leo, line 5: I created this dedicated function because using a single `isScrollable` function containing both checks doesn't work for the following reasons:

- `isScrollingEnabled` needs to check if scrolling was explicitly enabled.
- `isScrollingDisabled` needs to check if scrolling was explicitly disabled.

Combining both into `Boolean(isScrollingEnabled && !isScrollingDisabled)` (in a single `isScrollable` function) would imply the element won't be considered scrollable unless it explicitly had scrolling enabled, which is not something that is wanted when checking `node.scrollableElement` in `getScrollableAncestors.ts` because that element should always be considered scrollable, unless scrolling is explicitly disabled.

clauderic:

Issue #857, opened 2022-08-04
Cancel drag when the the draggable item loses focus


marcospassos:
Currently, when dragging using the keyboard, the item remains active even when focusing out of it. It leads to [several issues](https://github.com/ariakit/ariakit/issues/1684#issuecomment-1200007524) that can't be solved at the userland.

I tried implementing a custom sensor, but the current design doesn't allow for extension (note the `ts-ignore`):

```ts
/**
 * A custom keyboard sensor cancels the drag when moving the focus out of the draggable element.
 */
class CustomKeyboardSensor extends KeyboardSensor {
    private readonly sensorProps: SensorProps<SensorOptions>;

    constructor(props: KeyboardSensorProps) {
        super(props);

        this.sensorProps = props;

        this.handleDragCancel = this.handleDragCancel.bind(this);

        this.attachSensor();
    }

    private attachSensor(): void {
        this.sensorProps.event.target?.addEventListener('blur', this.handleDragCancel);
    }

    private detachSensor(): void {
        this.sensorProps.event.target?.removeEventListener('blur', this.handleDragCancel);
    }

    private handleDragCancel(): void {
        // There's no other way to notify the parent class to clean up the state.
        // @ts-ignore
        this.detach();
        this.detachSensor();
        this.sensorProps.onCancel();
    }
}
```

Even ignoring the error, other issues arise. For example, using the `CustomKeyboardSensor`, when pressing enter to finish dragging, DND kit announces the event as a cancel operation. Even worse, after calling the `onCancel` callback, in response to a click outside, the item steals the focus back, so you need to click twice to move the focus away from the item.

codewitch:
Would love assistance on this as well. If a non keyboard user accidentally gets into a keyboard drag state, the app will feel like it is broken since they cant cancel the drag by clicking elsewhere.

One thought would be to build a combination sensor that taps into both keyboard and click events?

pwang347:
For what it's worth, here's something hacky I did that seems to be working fine and not running into the issues mentioned above. Note also that in my application I am using `handleEnd` instead of `handleCancel`.

```typescript
export class CustomKeyboardSensor extends KeyboardSensor {
    constructor(props: KeyboardSensorProps) {
        super(props);

        const handleEnd = (this as any).handleEnd.bind(this);
        const detach = (this as any).detach.bind(this);

        props.activeNode.activatorNode.current?.addEventListener("blur", handleEnd);

        // hack: override the detach method to also remove the blur event
[...]

PR #1087, opened 2023-03-28
Deactivate keyboard sensor on blur

jperals:
This fixes https://github.com/clauderic/dnd-kit/issues/857.

As stated in that issue, this cannot be fixed by extending the KeyboardSensor, because all methods are private and this also creates other problems —but adding these changes in the main KeyboardSensor class works as expected. Moreover, I think it makes sense to deactivate dragging on blur as a general case in order to keep the UX consistent with the native browser focus —otherwise this leads to an inconsistent experience where a non-focused element keeps being interactive, which hurts usability and accessibility.

This works for losing focus in any way, therefore both by clicking with the mouse somewhere else or by typing Tab, even though in the latter case the (native) focus will stay on the active draggable element and a new Tab is required to navigate to the next focusable element, because of `RestoreFocus`. But since that can be overridden by setting `restoreFocus` to `false` in the `DndContext`, I did not include more changes in this PR in order to keep it minimal.

### DIFF
--- a/.changeset/silver-games-cough.md
+++ b/.changeset/silver-games-cough.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/core': patch
+---
+
+Deactivate keyboard sensor on blur
--- a/packages/core/src/sensors/events.ts
+++ b/packages/core/src/sensors/events.ts
@@ -1,4 +1,5 @@
 export enum EventName {
+  Blur = 'blur',
   Click = 'click',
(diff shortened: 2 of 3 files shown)


barrymichaeldoyle:
The keyboard sensor is a real mess in its current state in my opinion. Everything breaks if the user tabs away while dragging an element. Hopefully this PR solves that issue.

clauderic:
Closing this in favor of #1555

PR #1088, opened 2023-03-30
useDraggable - render only dragging item

alissaVrk:
in this PR we solve only the re-rendering for `useDraggable` hooks (and the components using it) when active element changes.

did not try to solve similar issues (yet) when `over` changes or for `useSortable` which exposes `activeIndex` and `overIndex` - to solve the issue there we will need to find a different API.
Maybe something like over: 'item-before' | 'me' | 'item-after' | 'other' so that the whole list won't re-render while dragging.

I tried to make as little changes as possible, but there are still some pretty significant changes, including changes in the API:
- `useDraggable` not returns `null` for `active`, `activatorEvent`, `activeNodeRect` if item isn't dragging
- `useDroppable` does not expose `active`.

please review and if you like the direction give some feedback. 
if you like the direction I'll see if this PR needs some fine-tuning and start thinking about the same issue in other areas.

### DIFF
--- a/cypress/integration/draggable_spec.ts
+++ b/cypress/integration/draggable_spec.ts
@@ -416,4 +416,17 @@ describe('Draggable', () => {
         });
     });
   });
+
+  describe('Multiple Draggables', () => {
+    it('should render only dragging element', () => {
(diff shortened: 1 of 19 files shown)


clauderic:

Issue #1092, opened 2023-04-01
Auto scroll is broken when using multiple containers


akantic:
First I want to give some praise to this repo, it's nothing short of amazing. Code is really neat with plenty to learn from and it is a really powerful tool. Thanks!

Nevertheless, I've found one other issue (besides experiencing this one - #1080), and now it's about the auto scroll feature.

If we take the official example from the story book ([Multiple Containers - Basic Setup](https://master--5fc05e08a4a65d0021ae0bf2.chromatic.com/?path=/story/presets-sortable-multiple-containers--basic-setup)), and we limit the height of horizontally sortable containers and make their `overflow-y` to `auto` or `scroll`, then horizontal auto scroll becomes basically unusable.

Here is a video where I copied the code from this repo to recreate the same example from the storybook with the additions mentioned above. Notice how it breaks on empty containers, while it seems to work fine when there are items in them. I tried conditionally adding `overflow-y` property if there are multiple children present, but then it happens for non-empty containers and not for the empty ones.

https://user-images.githubusercontent.com/20325132/229306611-35b95d23-ed84-4fe4-9f2a-b66aa8d1e07f.mov

Literally just removing `overflow-y` fixes the jitter, but it would be neat if we would be able to scroll even vertically. 

I already tried altering the `autoScroll` config:
- disabling `layoutShiftCompensation` does not help
- removing y axis scrolling does not help ( e.g. `threshold: { x: 0.2, y: 0 }` )
- added a `canScroll` function to skip the nested containers from auto scrolling vertically and only allow scrolling horizontally in the parent container - does not work

Any help is appreciated, thanks!

andelkocvjetkovic:
I'm facing the same issue, did you find a solution ?

andelkocvjetkovic:
I found this, https://github.com/clauderic/dnd-kit/issues/1108. It solved my problem

akantic:
My solution is the linked PR above, I am using a patched local package of this repo.

toss-fe-interview:
[#1094](https://github.com/clauderic/dnd-kit/pull/1094) works well.
You saved my time thank you :)

yungo1846:
https://github.com/clauderic/dnd-kit/pull/1094 works well.
You saved my time thank you :)

clauderic:
This issue was opened against a previous version of dnd-kit. The library has since undergone a major rewrite.

[...]

PR #1094, opened 2023-04-03
Fix rects measuring on elements change

akantic:
A fix for #1092, I am not sure if there are any other implications with this change so any input is appreciated. 

#### What was happening?
When using multiple containers `scrollableAncestors` would get updated, but their rects would not. 
So, let's imagine dragging from one container to another, first we have something like:
```
scrollableAncestors = [parent, grandparent]
scrollableAncestorsRects = [parentRect, grandparentRect]
```
Then ancestors get updated to just one element in between switching containers, but rects **do not**.
```
scrollableAncestors = [grandparent]
scrollableAncestorsRects = [parentRect, grandparentRect]
```
Then, this code brings chaos:
https://github.com/clauderic/dnd-kit/blob/5c58f0fe5d19b5aaa5cf93572f3435f4a0a6e54f/packages/core/src/hooks/utilities/useAutoScroller.ts#L121-L122

Because for grandparent `scrollContainer` it would use index 0 to access rects, leading to `scrollContainerRect` being `parentRect` for grandparent container - causing seemingly random jumps as can be seen in the video from the issue referenced on top.

### DIFF
--- a/packages/core/src/hooks/utilities/useRects.ts
+++ b/packages/core/src/hooks/utilities/useRects.ts
@@ -30,8 +30,8 @@ export function useRects(
       elements.forEach((element) => resizeObserver?.observe(element));
     } else {
       resizeObserver?.disconnect();
-      measureRects();
     }
+    measureRects();
   }, [elements]);
 
   return rects;


Rieranthony:
Can we have thus merged? It's a bug causing the autoscroll to be junky and unstable

clauderic:
Closing in favor of https://github.com/clauderic/dnd-kit/pull/1543

PR #1162, opened 2023-06-27
Feature/measure droppables when idle

heikomat:


heikomat:
i'm very sorry, i clicked on the wrong base repository

Issue #391, opened 2021-07-31
DragEndEvent is coming as null on onDragEnd


yash1200:
Hey! I implemented the dnd feature on my app, the interface is like this:

![Screenshot 2021-08-01 at 1 13 38 AM](https://user-images.githubusercontent.com/38391570/127750950-e3c40b33-9397-48a8-bf97-350621f346bd.png)

The section on the left contains two elements in a column both wrapped with `Droppable`
and the section on the right contains the grid of elements (all wrapped with `Draggable`)

The issue is sometimes for the last row of the grid, when I drag the element and drop it over the Droppable element, the `over` is returned `null`. The code for `handleDragStart`, `handleDragEnd`, `Draggable` and `Droppable` are below for reference.

```ts
const handleDragStart = (event) => {
  setActiveId(event.active.id);
  let index = duplicates.findIndex((i) => i.name == event.active.id);
  if (index >= 0) {
    updateSelected(index);
  }
};

// The over is null sometimes
const handleDragEnd = ({ over }) => {
  setActiveId(null);
  if (over) {
    if (over.id === "list-drop") {
      addToList();
    } else if (over.id === "image-drop") {
      updateCurrent();
    }
  }
};
```
```ts
const Draggable = (props) => {
  const { attributes, listeners, setNodeRef } = useDraggable({
    id: props.id,
  });

  return (
    <div
      ref={setNodeRef}
      {...listeners}
      {...attributes}
      style={{ outline: "none" }}
    >
      {props.children}
    </div>
  );
};

const Droppable = (props) => {
  const { setNodeRef } = useDroppable({
    id: props.id,
  });

  return (
    <div ref={setNodeRef} style={props.style}>
      {props.children}
    </div>
  );
};

```

The code for the right section
```ts
<MaterialGrid container spacing={3}>
  {Object.assign([], duplicates).map((photo, i) => {
    return (
      <MaterialGrid item xs={grid ? 6 : 12} key={i}>
        <Draggable key={photo.name} id={photo.name}>
          <div
            id={(i + 100000).toString()}
            style={{
              cursor: "pointer",
              opacity: activeId == photo.name ? "0" : "1",
            }}
          >
            <RelatedDialogItem
              image={photo}
              selected={i === duplicateSelected}
            />
          </div>
        </Draggable>
      </MaterialGrid>
    );
  })}
</MaterialGrid>
<DragOverlay>
  {activeId ? (
[...]

PR #1167, opened 2023-06-29
Fix issue with over being null in onDragEnd

adrianmxb:
I am currently implementing drag and drop behavior within our application and I've noticed that sometimes the `over` property is set to null. Found a issue for this one as well: https://github.com/clauderic/dnd-kit/issues/391

After some investigation, I was able to reproduce this on storybook in the `MultipleContainers` example as well, as I first thought I made a mistake in the implementation. Its very hard to reliably reproduce it so pardon me for not providing a codesandbox. I tried writing a test for this but it turns out this is not very easy to achieve.

Ways on how to reproduce (takes quite some time):
1. Do a very fast click (mouse down, mouse up) on a `Draggable`.
2. The `over` property you get through `onDragEnd` is sometimes `null`.

It really feels like a timing issue, since the `sensorContext.over` is set through this `useIsomorphicLayoutEffect`:
https://github.com/clauderic/dnd-kit/blob/5c58f0fe5d19b5aaa5cf93572f3435f4a0a6e54f/packages/core/src/components/DndContext/DndContext.tsx#L579-L613

By setting `over` directly in the `sensorContext` instead of relying on the `useIsomorphicLayoutEffect` to fire before the `onDragEnd` callback is called.

### DIFF
--- a/.changeset/soft-mails-check.md
+++ b/.changeset/soft-mails-check.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/core': patch
+---
+
+Fix issue with over being null in onDragEnd
--- a/packages/core/src/components/DndContext/DndContext.tsx
+++ b/packages/core/src/components/DndContext/DndContext.tsx
@@ -567,6 +567,7 @@ export const DndContext = memo(function DndContext({
       };
 
       unstable_batchedUpdates(() => {
+        sensorContext.current.over = over;
         setOver(over);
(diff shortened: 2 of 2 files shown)


balexand:
I'm also having problems with `over` being `null` in `onDragEnd` due to an inconsistently reproducible timing issue. This pull-request does not fix the issue and doesn't seem like the correct approach.

PR #1196, opened 2023-07-22
Fix weird issue in arrayMove

### LINKED ISSUE #1195: Sortable arrayMove doesn't work as intended
Hello folks! I don't know why but when using `arrayMove`, sometimes a different item changes its position. Here's an example:

As you can see, when I move B, A moves too. [codesandbox link](https://codesandbox.io/p/sandbox/dnd-kit-weird-arraymove-behavior-7vxhyg?file=%2Fsrc%2FApp.tsx%3A55%2C31)

When I use this function instead of `arrayMove` it works as intended:
```ts
setFoos((items) => {
  const newItems = [...items];

   const activeIdx = newItems.findIndex((x) => x.id === active.id)!;
   const overIdx = newItems.findIndex((x) => x.id === over.id)!;

   [newItems[activeIdx], newItems[overIdx]] = [newItems[overIdx], newItems[activeIdx]];

   return newItems;
});
```

Am I missing something?

natan-fernandes:
fixes #1195

### DIFF
--- a/packages/sortable/src/utilities/arrayMove.ts
+++ b/packages/sortable/src/utilities/arrayMove.ts
@@ -3,11 +3,6 @@
  */
 export function arrayMove<T>(array: T[], from: number, to: number): T[] {
   const newArray = array.slice();
-  newArray.splice(
-    to < 0 ? newArray.length + to : to,
-    0,
-    newArray.splice(from, 1)[0]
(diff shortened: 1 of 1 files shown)


natan-fernandes:
Just realized that I should've used arraySwap instead

PR #1229, opened 2023-09-07
Use `aria-live="polite"` for announcements

ayy-bc:
When playing around with dnd-kit, I noticed that the `aria-live` property is set to "assertive" by default. I believe that "assertive" should only be used for time-sensitive/critical notifications that absolutely require the user's immediate attention (from: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions#live_regions). 

Also, changing it to "polite" makes it such that the announcement is made when the user is idle. I think that really helps with screen-reader not announcing each time a user moves the item (gets a little annoying when say I want to drag it 5 rows below, it keeps announcing for each row but doesn't complete the announcement as each time I press down, it announces the new position without the last one completing) and instead, announces it when the user is a little sure of where the item should be (i.e be idle). 

Also moved the `visuallyHidden` directly within the component and not as a global constant as this reduces the number of declared constants and limits the scope of visuallyHidden (since it is only used within `LiveRegion` anyways)

### DIFF
--- a/packages/accessibility/src/components/LiveRegion/LiveRegion.tsx
+++ b/packages/accessibility/src/components/LiveRegion/LiveRegion.tsx
@@ -3,10 +3,10 @@ import React from 'react';
 export interface Props {
   id: string;
   announcement: string;
-  ariaLiveType: "polite" | "assertive" | "off";
+  ariaLiveType?: "polite" | "assertive" | "off";
 }
 
-export function LiveRegion({id, announcement, ariaLiveType = "polite"}: Props) {
+export function LiveRegion({id, announcement, ariaLiveType = "assertive"}: Props) {
   // Hide element visually but keep it readable by screen readers
   const visuallyHidden: React.CSSProperties = {
     position: 'fixed',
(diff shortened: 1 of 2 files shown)


ayy-bc:
@clauderic just wanted to bring this to your attention. Really appreciate you and this library, thank you so much.

packages/accessibility/src/components/LiveRegion/LiveRegion.tsx
```diff
@@ -3,29 +3,30 @@ import React from 'react';
...
-  border: 0,
-  padding: 0,
-  overflow: 'hidden',
-  clip: 'rect(0 0 0 0)',
-  clipPath: 'inset(100%)',
-  whiteSpace: 'nowrap',
-};
-
-export function LiveRegion({id, announcement}: Props) {
+export function LiveRegion({id, announcement, ariaLiveType = "polite"}: Props) {
```
hussam-i-am, line 9: @ayy-bc Maybe we should keep this defaulted to `assertive` so it doesnt affect any existing usages

hussam-i-am:
@ayy-bc Also, can you add a changeset to this? Thanks!

packages/accessibility/src/components/LiveRegion/LiveRegion.tsx
kendallgassner: I think assertive would be the way to go! -- User research indicated that this woul

clauderic (approved):

PR #1321, opened 2024-01-15
Add `initial` and `delta` to `pointerCoordinates` in `collisionDetection`

someonewithpc:
This was done this way to avoid breaking existing usages, but

```
export interface DragCoordinates {
  initial: Coordinates;
  current: Coordinates;
  delta: Coordinates;
};
```

would perhaps be preferable.

My motivating usage is to allow considering the center of the `active` rather than the `pointerCoordinates`

### DIFF
--- a/.changeset/poor-coats-compete.md
+++ b/.changeset/poor-coats-compete.md
@@ -1,0 +1,11 @@
+---
+'@dnd-kit/core': minor
+---
+
+Add `initial` and `delta` to `pointerCoordinates` in `collisionDetection`
+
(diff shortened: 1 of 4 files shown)


someonewithpc:
> This PR includes changesets to release 3 packages
> Name	Type
> @dnd-kit/core	Minor
> @dnd-kit/modifiers	Major
> @dnd-kit/sortable	Major

I don't know why it's detecting changes in the other two packages, as the changeset file starts with:

```
---
'@dnd-kit/core': minor
---
```

PR #1376, opened 2024-03-28
Shallow copying the entire object to copy owned as well as prototype properties

Rohitbels:
Spread paramters will only copy enumerable and own properties from a source object to a target object, 

you are able to use this method and [Object.create()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create) for a [shallow copy](https://developer.mozilla.org/en-US/docs/Glossary/Shallow_copy) between two unknown objects.


Rohitbels:
@clauderic 
Can we please review and merge this?

Rohitbels:
My Custom data is an Object with properties inherited from Prototype.
This wont break any existing stuff.

PR #1448, opened 2024-07-03
[Experimental branch] Allow the entity to have a new id

lfades:
When the `id` of the entity is updated from a component I noted that the element wasn't draggable anymore, as it if were not being detected. The reason was that the new `id` was not being compared in the `id` effect at `Entity`.

What I did was to change the effect to read the `id` from `this` rather than using the `id` received by the `constructor`. This makes sense because the new `id` is set to the class in the hooks:

```
useOnValueChange(id, () => (droppable.id = id));
```

With this change it should be straightforward to add a "Clone" example to the stories. I can do that too if you're interested.

### DIFF
--- a/packages/abstract/src/core/entities/entity/entity.ts
+++ b/packages/abstract/src/core/entities/entity/entity.ts
@@ -38,7 +38,7 @@ export class Entity<
     this.effects = () => [
       () => {
         // Re-run this effect whenever the `id` changes
-        const {id: _, manager} = this;
+        const {id, manager} = this;
 
         if (id === previousId) {
           return;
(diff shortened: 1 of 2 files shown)


packages/abstract/src/core/entities/entity/entity.ts
```diff
@@ -38,7 +38,7 @@ export class Entity<
     this.effects = () => [
       () => {
         // Re-run this effect whenever the `id` changes
-        const {id: _, manager} = this;
+        const {id, manager} = this;
```
clauderic, line 41: What's the difference between this code and the previous code?

aarjithn: weren't we discarding the value of id from this here? (by assigning to _)

lfades: The difference is when the effect runs again, it will use whatever id is set in `this`, instead of the `id` that was received by the class constructor. That way if the id changes you're using the new id, not the initially set one.

Issue #1108, opened 2023-04-20
React 18: Autoscroll is broken with multiple horizontal drop zones which have vertical scroll


ThomDevine:
Hi, 

I'm seeing this behavior only after updating to React 18.  I was using 6.0.5, updating dnd-kit to 6.0.8 does not resolve this.

### Setup
I have a horizontal, scrollable list of "useDroppable" columns. Each column contains a vertical list of "useDraggable" items.  When the column contents' CSS **overflow** is set to **auto**, **scroll**, or **hidden**, this behavior occurs. If I change it to **none**, scrolling works as it did previously.

### Behavior
When I try to drag an item from one column to another, as soon as the item is over a different drop zone, the main container (of columns) continues to scroll horizontally all the way to the end.  I cannot drag in the other direction to scroll it back.

Changing DndContext's **autoScroll** with various options of **order** and **acceleration** has no effect.

### Sandbox Eample
https://codesandbox.io/s/stoic-robinson-wgcrcx?file=/src/App.jsx
There are no handlers attached to DndContext, no underlying data is changed, so this appears to be contained within autoScroll.

Try dragging from a middle column to see the "unable to reverse" behavior. Search for "changeme" in App.js to find the relevant style change in the Column component. In my app, this style is set by a CSS file instead of inline, and the behavior is the same.

Note: I'm not using Sortable here because the column contents have a strict order defined by the server. https://github.com/clauderic/dnd-kit/issues/1098 might be related, but that's for Sortable so I'm not sure.

ThomDevine:
I have confirmed that by loading my app the old way, with render(app, mountPoint) from the "react-dom" package to get React 17 behavior, this bug goes away.  This is not a valid workaround though, as this method is [not supported](https://react.dev/blog/2022/03/08/react-18-upgrade-guide#updates-to-client-rendering-apisl) in React 18.

csc-bo:
I encountered the same issue. Does anyone have a solution for React 18?

alexdonets:
Same here. @ThomDevine have you found a workaround yet?

ThomDevine:
@alexdonets No, we are still stuck on React 17 because of this issue.

I poked around in the library code to try and figure it out, which is how I came across the **order** option for autoScroll; it's not found in the docs.  I expected setting that to `TreeOrder.ReversedTreeOrder` to work, that _seems_ like what it's designed for, but it didn't.

[...]

Issue #1080, opened 2023-03-24
MultipleContainers jitter from inner item when dragging to bottom of new column


brettinternet:
I have reconstructed the multiple containers example:

https://codesandbox.io/s/dndkit-ivpj4d

When moving an inner container to the bottom of a different column, I'm experiencing some jitter from an existing inner item at the bottom of the column just as the dragged item is carried to the new column.

Watch "C4" in this video:

https://user-images.githubusercontent.com/20871604/227420071-22394f93-86d4-45d0-8b90-198850516e71.mov

<br />

In a local project:

https://user-images.githubusercontent.com/20871604/227420320-4a7f096e-3731-4c87-ad1d-23e76bcb2f4d.mov

<br />

It appears subtle in this video, but it appears more pronounced in different examples. Does anyone know what might be causing this jitter?

akantic:
I noticed the same behaviour when I've added more logic to my draggable items.

I did some debugging and I think what happens is that the vertical sorting strategy first adds the item as next to last, then quickly moves it to the last spot when the state updates with the item in new container. As of now, I have no solution except to try and simplify my draggable items when dragging something.

In this video, first I have the full item and you can see how it jitters when changing containers. Then, I comment out the menu code (button with 3 vertical dots, which contains plenty of logic) inside the draggable item and the jitter is gone. 

[](https://user-images.githubusercontent.com/20325132/228883736-2c377f36-fe80-4b04-8b2b-7f7db356144b.mp4
)

brettinternet:
I think you're right, it's probably related to the vertical sort or the collision strategy. I also noticed the keyboard's programmatic move will move the item to the next-to-last in the next column as well, and refuses to move it to the last item, even if it's lower on the previous column's list.

clauderic:
This issue was opened against a previous version of dnd-kit. The library has since undergone a major rewrite.

It has been rebuilt from the ground up with support for **multiple frameworks** (React, Vue, Svelte, Solid, and vanilla JavaScript), improved performance, and a new API.

- **Documentation**: https://dndkit.com
- **Migration guide**: https://dndkit.com/react/migration
- **GitHub**: https://github.com/clauderic/dnd-kit

If this issue is still relevant to the latest version, please open a new issue.

Issue #901, opened 2022-09-30
DragOverlay misplaced when sorting multiple elements 


tozz:
I've used the multi column multi sort example (https://github.com/clauderic/dnd-kit/pull/588) but I can't figure out how to actually get the overlay to get to the right position when removing more than one node from the list, as happens in the storybook.

https://user-images.githubusercontent.com/705677/193299052-33b7ceec-da18-4589-a25e-ea6d7c2e002e.mov

I made a minimal example here https://github.com/tozz/dnd-multisort where you can `yarn start` to test in your browser, the files are in `src/js/` and `component.tsx` is the main entry point for the DnD stuff. I tried to keep the example as minimal as possible.

tozz:
> I am having the same problem. Were you able to solve this issue?

Unfortunately not, I moved on to other things but it would be great if someone has a fix :)

carlosloureda:
@tozz Thanks. I just found that the example at #588 was using outdated versions of the packages:

`@dnd-kit/core@5.0.1 @dnd-kit/sortable@6.0.0 @dnd-kit/utilities@3.1.0`

Using those versions seem to fix the problem of the cursor position. But now some things (like moving into another container are broken on my example).

megos:
Not so a good code, but I come up with a workaround.
It is how to delay updating the state on `onDragStart`.

In this example, delay updating `filteredItems` to prevent the number of items from changing.
This will help calculate the correct position.

```js
onDragStart={e => {
  setTimeout(() => {
    doSomething()
  }, 1)
}}
```

kamilkazmierczakMtab:
Did someone find other solution to this because using setTimeout creates a lot of other issues and wrapping it with promise does not solve problem because delaying whole events like `onDragEnd` until execution of `onDragStart` breaks dnd behavior and delaying only parts of it creates yet another issues.

maloguertin:
it would be nice to have a onBeforeDragStart where it is safe to cause rerenders without messing with the position.

linxianxi:
same. I don't think the cursor position is wrong, the overlay should be at the current mouse dragging location.

williamisnotdefined:
I'm facing the same problem, fortunately the setTimeout did work for me.

x-Foz3R-x:
Got the same issue with my horizontal list dragOverlay and i have pretty simple workaround that just shifts overlay to correct position :)

Keep in mind that this is example for horizontal list, but its very simple to modify it for the vertical list :P

```typescript
[...]

PR #1461, opened 2024-07-28
fix: auto-scroll bug with multiple containers

kswenson:
In April 2023, `akantic` identified an auto-scroll issue with `dnd-kit` in #1092 and then proposed a fix for it in #1094. In September 2023, unaware of `akantic`'s PR, I fixed the same issue in CODAP in [CODAP #901](https://github.com/concord-consortium/codap/pull/901) with a different local patch that has similar effect. In comparing the two different fixes, I now believe that `akantic`'s fix is better than the one that I came up with, but that it can still be improved upon, as demonstrated in this PR.

The code in question comes from the `useRects()` hook, which is responsible for maintaining/returning an array of rectangles that correspond to the bounding rectangles of an array of elements passed in as an argument. (In the context in which it is used, this array of elements is returned from a call to `useScrollableAncestors()`.) The code in question is:

```typescript
  if (elements.length > 0 && rects === defaultValue) {
    measureRects();
  }

  useIsomorphicLayoutEffect(() => {
    if (elements.length) {
      elements.forEach((element) => resizeObserver?.observe(element));
    } else {
      resizeObserver?.disconnect();
      measureRects();
    }
  }, [elements]);
```

where `measureRects()` is the dispatcher returned from a call to the `useReducer()` hook that is responsible for updating the state to the correct array of bounding rectangles. The current code guarantees that `measureRects()` will be called on transitions to/from an empty `elements` array. The bug occurs when, for instance, on subsequent calls to the `useRects()` hook the `elements` array changes from one non-zero length to another. In this case, `measureRects()` is not called and so the returned array of bounding rectangles has the wrong length. In the words of #1094, "Then, this code brings chaos," pointing to the `useAutoScroller()` hook which indexes into the array of rectangles.

`akantic`'s fix was to move the call to `measureRects()` in the `useIsomorphicLayoutEffect()` hook outside of the `else` to the root of the effect so that it would be called whenever the hook ran, independent of whether the `elements` array was empty or not. This makes sense, given that any time the array of elements changes, we need to update the array of bounding rectangles.

My original fix was to change the conditional that precedes the effect to:

```typescript
  if (elements.length !== rects.length) {
    measureRects();
  }
```

This has similar effect, in that `measureRects()` is called any time the length of the `elements` array changes, but in this case it is called at the root of the render function outside of any effects. While this fixes the issue we were encountering in which the length of the `elements` array was changing, a limitation of this fix is that it doesn't handle the case in which the `elements` array changes without changing length, which is why `akantic`'s fix is preferable.

Neither of these earlier fixes accounts for the fact that the other purpose of the `useIsomorphicLayoutEffect()` hook is to connect/disconnect a resize observer to/from each of the elements in the array, however. Thus, with either of our fixes it is possible for the code to continue to observe elements that are no longer present in the `elements` array.

This PR implements what I believe to be an improved fix which replaces the original code with the simpler:

```typescript
  useIsomorphicLayoutEffect(() => {
    resizeObserver?.disconnect();
    measureRects();
    elements.forEach((element) => resizeObserver?.observe(element));
  }, [elements]);
```

The transition to/from an empty `elements` array is no longer a special case. Any time the `elements` array changes, we must disconnect the resize observer from the previous elements, call `measureRects()` to update the current set of bounding rectangles, and make sure that the current elements are observed. Doing so consistently inside the `useIsomorphicLayoutEffect()` obviates the need for the separate conditional call to `measureRects()` outside the effect.

See also:
- #1092 -- `akantic`'s original issue
- #1094 -- `akantic`'s original PR
- #1080, #1108 -- potentially related, but I didn't investigate further
- [CODAP #901](https://github.com/concord-consortium/codap/pull/901) -- CODAP's original patch PR

### DIFF
--- a/packages/core/src/hooks/utilities/useRects.ts
+++ b/packages/core/src/hooks/utilities/useRects.ts
@@ -21,17 +21,10 @@ export function useRects(
   const [rects, measureRects] = useReducer(reducer, defaultValue);
   const resizeObserver = useResizeObserver({callback: measureRects});
 
-  if (elements.length > 0 && rects === defaultValue) {
-    measureRects();
-  }
-
   useIsomorphicLayoutEffect(() => {
-    if (elements.length) {
-      elements.forEach((element) => resizeObserver?.observe(element));
-    } else {
-      resizeObserver?.disconnect();
-      measureRects();
-    }
+    resizeObserver?.disconnect();
+    measureRects();
+    elements.forEach((element) => resizeObserver?.observe(element));
   }, [elements]);
 
   return rects;


clauderic:
PR was missing a changeset and I don't have permissions to push to your branch, re-opening here https://github.com/clauderic/dnd-kit/pull/1543

kswenson:
> PR was missing a changeset and I don't have permissions to push to your branch, re-opening here #1543

Sorry about the missing changeset. 😬

clauderic:
np, thanks for the comprehensive PR description 👍

Issue #1379, opened 2024-04-01
useSortable re-renders all items even with just clicking one item, affecting performance


wdire:
I noticed a lag while moving a sortable item on mobile when having around 50 sortable items. I've already noticed that when even only clicking a item it does render every other items, idk why. So I moved everything beside `useSortable` to `ContentCardMemo` that is wrapped with `React.memo`, that `memo` component doesn't re-render every time and I _thought_ its ok now. I've made sure that `useSortable` is the one causing all re-renders, the container that maps all `ContentCard`s is not the one re-rendering.

It might have improved with `memo` but I'm having performance issues on mobile with a merely 50 items. First I thought my way of storing/updating items might have causing the issue, storing it in a array of objects, when order changes it recreates whole items with `arrayMove`. But considering container not being re-rendered and just touching a sortable item causes total 200 re-renders, the issue is about `useSortable` itself.

Though on desktop it's not as bad as mobile, it still shows lag, performance issue when the list grows more.

Is there a way to fix this?

Versions:
"@dnd-kit/core": "^6.1.0",
"@dnd-kit/modifiers": "^7.0.0",
"@dnd-kit/sortable": "^8.0.0", // also tried with 7.0.2, same
"@dnd-kit/utilities": "^3.2.2",

Code at issue:
```
// ContentCard.tsx

const ContentCard = memo(function ContentCard({content}: Props) {
  const redirectSourcePage = useAppSelector((state) => state.list.redirectSourcePage);

  const {setNodeRef, attributes, listeners, transform, transition, isDragging} = useSortable({
    id: content.id,
    data: {
      type: "Content",
    },
    disabled: redirectSourcePage,
  });

  // This runs 4 times each. If there are 50 items, 200 times even with just touching to any item.
  console.log("card", content.id);

  return (
    <div
      ref={setNodeRef}
      style={{
        transition,
        transform: CSS.Translate.toString(transform),
      }}
      {...attributes}
      {...listeners}
      data-contentcard="true"
    >
      <ContentCardMemo
        content={content}
        redirectSourcePage={redirectSourcePage}
        isDragging={isDragging}
      />
    </div>
  );
});

export default ContentCard;
```

lukjaki:
Encountered this but got solved by wrapping whole component in memo() where the second argument is the function that checks if component should rerender.

wdire:
[...]

PR #1494, opened 2024-09-29
Avoid re-rendering children on click event

dinkinflickaa:
I noticed that clicking on a DnD child node can be costly because it triggers all child nodes to rerender twice. This issue has been reported before. See #1379 

This happens when `DndContext` updates the `activeSensor` and `activatorEvent` states on a mouse down event. These state changes cause rerenders in all consumers of `DndContext`.  Consumers of `useSortable` re-render too because `SortableContext` depends on `DndContext`.

If the user is just clicking or if activation constraints aren’t met, DndContext resets these states to null, causing another render cycle.

Here’s what it looks like in React Profiler for a click event, where you'll see back-to-back commit cycles causing unnecessary rerenders for every child:
image

I don't see why we need to update state unless the user is actually dragging. This PR proposes delaying the` activeSensor` and `activatorEvent` state updates until the drag event starts and constraints are met. This reduces unnecessary re-renders, saving two render cycles for every child.

Here’s what the React Profiler looks like with this change:

image

From Perf trace, you'll notice that mouse down and mouse up events aren't doing much work
Before:
image

After: 

image

Cypress tests are passing -

image

### DIFF
--- a/.changeset/brave-kids-explain.md
+++ b/.changeset/brave-kids-explain.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/core': patch
+---
+
+Improves performance by eliminating wasteful re-renders on every child item on click
--- a/packages/core/src/components/DndContext/DndContext.tsx
+++ b/packages/core/src/components/DndContext/DndContext.tsx
@@ -320,7 +320,6 @@ export const DndContext = memo(function DndContext({
   );
 
   const activeSensorRef = useRef<SensorInstance | null>(null);
-  const activatorEventRef = useRef<Event | null>(null);
   const instantiateSensor = useCallback(
     (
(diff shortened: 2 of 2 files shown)


alloy:
@dinkinflickaa Can you please run the e2e tests locally and show that they pass? Also, can the e2e tests be amended any way to protect against this regressing in the future? (I assume it would require more tooling, such as integrating React Render Tracker?)

dinkinflickaa:
@alloy Thanks for pointing out tests. This helped me uncover a bug which I have fixed now. Added a screenshot of passing tests to PR description.

SudoPlz:
We're experiencing the same issue in our project, any idea when this will be merged?

clauderic (approved):

PR #1499, opened 2024-10-10
Always set proxy

chrisvxd:
Fix a bug in the `experimental` branch that prevented all unique droppables that share an element from each receiving the cloned proxy.

This is because the underlying IDENTIFIER_ATTRIBUTE was getting overwritten for each droppable, and so only the last droppable item would have the correct identifier attribute and hence proxy.

Instead, generate a unique data attribute for each droppable using `generateUniqueId` (now moved out of the accessibility plugin) and store it in a map. Note, we can't reuse the element's ID, since this may contain illegal characters.

### DIFF
--- a/packages/dom/src/core/plugins/feedback/Feedback.ts
+++ b/packages/dom/src/core/plugins/feedback/Feedback.ts
@@ -15,18 +15,18 @@ import {
   isSafari,
   getWindow,
   type Transform,
+  generateUniqueId,
 } from '@dnd-kit/dom/utilities';
 import {Coordinates} from '@dnd-kit/geometry';
 
 import type {DragDropManager} from '../../manager/index.ts';
-import type {Draggable} from '../../entities/index.ts';
+import type {Draggable, Droppable} from '../../entities/index.ts';
 
 const ATTR_PREFIX = 'data-dnd-';
 const CSS_PREFIX = '--dnd-';
 const ATTRIBUTE = `${ATTR_PREFIX}dragging`;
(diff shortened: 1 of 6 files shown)


packages/dom/src/core/plugins/feedback/Feedback.ts
```diff
@@ -520,10 +521,12 @@ function createPlaceholder(source: Draggable) {
         element === droppable.element ||
         element.contains(droppable.element)
       ) {
-        droppable.element.setAttribute(
-          IDENTIFIER_ATTRIBUTE,
-          `${JSON.stringify(droppable.id)}`
-        );
+        const identifierAttribute = `${ATTR_PREFIX}${generateUniqueId('id')}`;
```
clauderic, line 524:
```suggestion
        const identifierAttribute = `${ATTR_PREFIX}${generateUniqueId('dom-id')}`;
```

packages/dom/src/core/plugins/feedback/Feedback.ts
```diff
@@ -512,6 +512,7 @@ function createPlaceholder(source: Draggable) {
 
     const {droppables} = manager.registry;
     const containedDroppables = [];
+    const identifiers = new Map<Droppable, string>();
```
clauderic, line 515: I think you could probably change `containedDroppables` to a `Map<Droppable, string>` and get rid of `identifiers`.

chrisvxd: Good idea. Updated 👍

clauderic (approved):

Issue #598, opened 2022-01-21
DragOverlay disappears if an item is selected before drop animation completes


repercussive:
👇 recording of interaction with the official examples

1. an item is picked up
2. while the `DragOverlay` is animating to its final position, another item is picked up
3. the `DragOverlay` does not appear for the second item.

![example](https://user-images.githubusercontent.com/7796522/150571719-53a9c5c7-fad6-4b7f-af00-afcccfb560c4.gif)

Given there is only ever 1 `DragOverlay` component at a time, I understand it wouldn't be possible for it to simultaneously animate to the old position while also appearing under the cursor. But **I would expect the overlay to always prioritize the active item**.

Only solution appears to be setting the `dropAnimation` duration to 0, circumventing the bug but sacrificing the nice animation.

PR #1500, opened 2024-10-10
Expose collision via notifier, and add data param

chrisvxd:
Make two changes to support advanced collision detection algorithms in the `experimental` branch:

1. Add the optional `data` parameter to the Collision type, to allow collision detector algorithms to provide metadata about the collision
2. Expose the `collision` via the notifier (on DataOperation), so the user can access the collision (including the new `data` parameter) in their drag callbacks

My use-case is a a sophisticated collision detection algorithm for Puck in #598 that provides [`position: "before" | "after"` data](https://github.com/measuredco/puck/pull/598/files#r1795882946) based on the intersection of the two items.

### DIFF
--- a/.changeset/add-data-param-to-collision.md
+++ b/.changeset/add-data-param-to-collision.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/abstract': patch
+---
+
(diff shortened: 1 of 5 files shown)


chrisvxd:
Closing as possible with dragOperation.data.

PR #1510, opened 2024-10-25
Add same-origin iframe support

chrisvxd:
Add some stories for same-origin iframe support, and patch the implementation to calculate the iframe position. This implementation works with both pointer and keyboard sensors.

The current implementation does not seem to behave well when `popover` is set. I've added a comment in the code to highlight the line, but `popover` is not necessary for my Puck implementation so I'd be happy to disable it if necessary.

I've paused here to see if I'm on the right track, or if I'm missing something.

## Tasks

- [x] Add stories
- [x] Modify pointer intersection
- [x] Offset shape in DOMRectangle
- [x] Update Feedback plugin
- [x] Fix popover implementation
- [x] Prevent popover from flashing in storybook
- [x] Add changeset
- [x] Add iframe scroll support to the scroller plugin
- [x] Add support for transformed iframes
- [ ] Account for parent document scrolling
- [x] Tidy stories
- [x] ~Fix bug where it's possible to drag items outside of lists in story examples~ Existing bug


apps/stories/stories/react/Sortable/iframe/IframeExample.tsx
```diff
@@ -0,0 +1,232 @@
...
+                  key={id}
+                  id={id}
+                  column={columns[1]}
+                  index={index}
+                  style={grid ? {height: 100} : undefined}
+                />
+              ))}
+            </SortableColumn>
+          </div>
+        </AutoFrame>
```
chrisvxd, line 141: This could definitely be tidied up

packages/dom/src/core/plugins/feedback/Feedback.ts
```diff
@@ -21,6 +21,7 @@ import {Coordinates} from '@dnd-kit/geometry';
 
 import type {DragDropManager} from '../../manager/index.ts';
 import type {Draggable, Droppable} from '../../entities/index.ts';
+import getIframeOffset from '../../../utilities/iframe/get-iframe-offset.ts';
```
chrisvxd, line 24: Should probably export these from utilities barrel

packages/dom/src/utilities/shapes/DOMRectangle.ts
```diff
@@ -19,10 +21,16 @@ export class DOMRectangle extends Rectangle {
...
     const resetAnimations = forceFinishAnimations(element);
-    let {top, left, right, bottom, width, height} =
-      getBoundingClientRect(element);
-    const computedStyles = getComputedStyles(element);
+    const iframeOffset = getIframeOffset(element);
+    const rect = getBoundingClientRect(element);
+    let {top, left, right, bottom, width, height} = applyOffset(
+      rect,
+      iframeOffset
+    );
```
clauderic, line 29: You can get rid of `applyOffset` and do this instead:

```suggestion
    let {top, left, right, bottom, width, height} = Rectangle.from(rect).translate(iframeOffset.x, iframeOffset.y);
```

clauderic (commented):
Looking very promising, just a couple of suggestions

packages/dom/src/utilities/execution-context/getIframe.ts
```diff
@@ -0,0 +1,13 @@
...
+  const win = getWindow(target);
+
+  if (win && win.self !== win.parent) {
+    const iframe = win.frameElement as HTMLIFrameElement | null;
+
+    return iframe;
+  }
+
+  return null;
+}
```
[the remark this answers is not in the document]
chrisvxd: Oops, yes!

packages/dom/src/core/plugins/feedback/Feedback.ts
```diff
@@ -205,8 +211,8 @@ export class Feedback extends Plugin<DragDropManager, FeedbackOptions> {
           {
             width: placeholderShape.width,
             height: placeholderShape.height,
-            top: top + dY + offset.top,
-            left: left + dX + offset.left,
+            top: top + dY + offset.top - iframeOffset.y,
+            left: left + dX + offset.left - iframeOffset.x,
```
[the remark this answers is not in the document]
chrisvxd: I think there are actually two related popover issues with this implementation:

1. A flash of the elements the centre of the frame, corrected as soon as the ResizeObserver callback triggers,  often too quickly to even render. This is  what I initially reported.
2. An incorrect position during the entire first drag, only noticeable when the frame is certain dimensions. Not sure why this is linked to the frame dimension, but it always goes away on subsequent drags.

Offsetting the _projected_ shape addresses (2). Offsetting the initial coordinates results in new issues. I'll continue to investigate.

packages/dom/src/core/plugins/feedback/Feedback.ts
```diff
@@ -52,6 +55,19 @@ export class Feedback extends Plugin<DragDropManager, FeedbackOptions> {
...
+        if (element) {
+          const elDoc = getDocument(element);
+
+          if (elDoc !== document) {
+            elDocStyle = getDocument(element).createElement('style');
+            elDocStyle.innerText = cssRules;
+            elDoc.head.prepend(elDocStyle);
+          }
+        }
+
```
chrisvxd, line 70: @clauderic this fixes the popover "flashing". dnd-kit needs to handle this rather than user since this needs to be injected before rendering the Feedback plugin.

clauderic: Can you refactor this to:

```ts
    const styleTags = new Map<Document, HTMLStyleElement>();

    const styleInjectionCleanup = effect(() => {
      const {status, source} = manager.dragOperation;

      if (status.initialized) {
        const doc = getDocument(source?.element ?? null);

        if (styleTags.has(doc)) return;

        const style = document.createElement('style');
        style.innerText = cssRules;
        doc.head.prepend(style);

        return styleInjectionCleanup;
      }
    });
```

And add this to the `this.destroy` method on line 521:

```ts
styleTags.forEach((style) => style.remove());
```

chrisvxd: This will prevent the styles from being injected into the host frame - the styles must be in both if dragging between the child and parent frames.

Although my original implementation handles that, your suggestion made me realise that it does _not_ account for the case when dragging between 2 iframes. Perhaps this needs further refactoring to dynamically inject styles into the document for the target element.

chrisvxd: I've pushed another change based on your suggestion, fixing the above issue: [f7b69d8](https://github.com/clauderic/dnd-kit/pull/1510/commits/f7b69d8df2536b9b7a7a9357342030ee7975dd3e)

There are 2 items outstanding in the PR description (scroller and tidying stories). Let me know if you'd like to do them as part of this, or separately.

packages/dom/src/core/plugins/feedback/Feedback.ts
```diff
@@ -47,10 +50,20 @@ export class Feedback extends Plugin<DragDropManager, FeedbackOptions> {
...
-        style = document.createElement('style');
-        style.innerText = cssRules;
-        document.head.prepend(style);
+      const {status, source, target} = manager.dragOperation;
+
+      if (status.initialized) {
+        const sourceDoc = getDocument(source?.element ?? null);
+        const targetDoc = getDocument(target?.element ?? null);
+
+        for (const doc of [sourceDoc, targetDoc]) {
```
clauderic, line 59: nit:

```suggestion
        const sourceDocument = getDocument(source?.element ?? null);
        const targetDocument = getDocument(target?.element ?? null);
        const documents = new Set([sourceDoc, targetDoc]);

        for (const doc of documents) {
```

packages/dom/src/core/plugins/feedback/Feedback.ts
```diff
@@ -171,17 +186,20 @@ export class Feedback extends Plugin<DragDropManager, FeedbackOptions> {
         showPopover(element);
       }
 
-      const actual = new DOMRectangle(element, {ignoreTransforms: true});
+      const actual = new DOMRectangle(element, {
+        ignoreTransforms: true,
+        ignoreFrameOffset: true,
```
clauderic, line 191: Do we really need `ignoreFrameOffset`? 🤔 I feel like my hacky prototype worked without it

chrisvxd:
Thanks for fixing the build @clauderic! I tested the latest in Storybook and Puck. Found that collisions are still not triggering as expected with transformed frames.

Here are some more concrete observations -

1. When dragging between transformed frames, the pointer coordinates are not scaled to match the transform, resulting in the dragged item lagging behind or running ahead of the cursor. This is why I was trying to fix it at the PointerSensor level before, but hopefully there's still a better way to do this. I believe the PointerSensor / Rect desync is causing the collision issues.
2. When dropping within the transformed frame, the end of the Feedback drop animation does not align with the final rendered position of the item, causing it to jump after the drop
3. When dragging between transformed frames, the drag handle location moves

clauderic:
Closing this PR and re-opening here: https://github.com/clauderic/dnd-kit/pull/1517

PR #1514, opened 2024-10-30
Iframe scroll

chrisvxd:
Another follow-on to #1510, which introduces support to offset the PointerSensor position when the iframe parent is scrolled. See 771d8a454cfc331ac308638e3e30a72e2761ec02.

This implementation is buggy, but wanted to share since we're working on similar areas.


packages/dom/src/core/sensors/pointer/PointerSensor.ts
```diff
@@ -258,6 +288,33 @@ export class PointerSensor extends Sensor<
...
+
+      // TODO this clashes with autoscroller
+      this.manager.actions.move({
+        to: {
+          x: this.coordinates.x + this.nestedScroll.x,
+          y: this.coordinates.y + this.nestedScroll.y,
+        },
+      });
+    }
+  }
```
clauderic, line 316: I think we can probably find a different way to solve this, most of this complexity can probably be handled in the [Scroller](https://github.com/clauderic/dnd-kit/blob/experimental/packages/dom/src/core/plugins/scrolling/Scroller.ts) plugin and in utilities like [getScrollableAncestors](https://github.com/clauderic/dnd-kit/blob/experimental/packages/dom/src/utilities/scroll/getScrollableAncestors.ts). There's some obvious bugs in both because they don't currently account for the presence of other frames, one example of this is [the Scroller plugin uses `document.elementFromPoint()`](https://github.com/clauderic/dnd-kit/blob/experimental/packages/dom/src/core/plugins/scrolling/Scroller.ts#L38-L44) but this returns the `iframe` element, and so we need to see if we can reach inside that iframe to find another element at that point.

My nudge here would be to add as little logic as possible to `PointerSensor`, its only responsibility should be to make sure that it reports coordinates that have the same point of reference as the main execution context.

chrisvxd: Makes sense! I wanted to push up earlier rather than later since we already discussed keeping things out of PointerSensor via Slack.

These draft PRs are first attempts at identifying and resolving various issues I've encountered, and really I'm just intending to figure out the bounds a bit and get the conversation going. More than happy to keep iterating and finding the right solution.

FWIW, I also realised I don't actually need this for Puck. But I do need the auto scroller to react to iframe scrolling.

chrisvxd:
Closing in favour of #1510

PR #1517, opened 2024-11-04
Add same-origin iframe support

clauderic:
Based on #1510

Add some stories for same-origin iframe support, and patch the implementation to calculate the iframe position. This implementation works with both pointer and keyboard sensors.


clauderic:
Going to go ahead and merge this @chrisvxd, I think it's working well enough to be included into the experimental branch and we can work on some improvements in separate PRs, as this one is getting quite large.

PR #1524, opened 2024-11-10
Add nested list stories

chrisvxd:
This PR:

1. adds nested list stories to reproduce a bug causing items to be positioned incorrectly when dragging between nested droppables. You can reproduce this bug by dragging an item out of the list and into the root.
2. fixes a bug where droppables can collide with their own child by tracking the path - I wasn't totally sure of the performance impact here, but I tried to minimise it by tracking it in the Droppable effect().
3. exposes several other issues with nested lists using the default dnd-kit behaviour (which Puck addresses with a plugin and custom collision algorithm).


packages/abstract/src/core/entities/droppable/droppable.ts
```diff
@@ -90,4 +95,6 @@ export class Droppable<
   public get isDropTarget() {
     return this.manager?.dragOperation.target?.id === this.id;
   }
+
+  public path: UniqueIdentifier[] = [];
```
clauderic, line 99: I would make this something like:

```
  @reactive
  public accessor parent: UniqueIdentifier | undefined;

  @derived
  public get path() {
    const {droppables} = manager.registry;
    const {parent} = this;
    const path = [];
    let parent = droppables.get(parent);
   
    while (parent) {
        path.push(parent.id);
        parent = droppables.get(parent.parent);
    }
    
    return path;
  }
```

And then the `@dnd-kit/dom` implementation can override `path` and automatically infer the path based on the DOM if `this.parent` is `undefined`

packages/dom/src/core/plugins/feedback/Feedback.ts
```diff
@@ -155,20 +154,23 @@ export class Feedback extends Plugin<DragDropManager, FeedbackOptions> {
...
+          (initialSize.height * initialFrameTransform.scaleY -
+            height * frameTransform.scaleY) *
           transformOrigin.y,
       };
+
       const delta = {
-        x: coordinatesDelta.x / frameTransform.scaleX - sizeDelta.width,
-        y: coordinatesDelta.y / frameTransform.scaleY - sizeDelta.height,
+        x: coordinatesDelta.x / frameTransform.scaleX + sizeDelta.width,
+        y: coordinatesDelta.y / frameTransform.scaleY + sizeDelta.height,
```
chrisvxd, line 171: Had to rework this logic to account for regular size changes, too.

It seems to be mostly working, but keyboard sensor positions item a few px out, which has me suspicious

clauderic:

Issue #1532, opened 2024-11-19
Cannot drag item with id={0} (number)


awerc:
If you define multiple draggable elements with numeric id, everything is working until you get element with id={0}, for that item onDragStart is triggered, but onDragMove is not, and that element cannot be moved

minimal reproduction example:
```javascript
import React from "react";
import { DndContext, useDraggable } from "@dnd-kit/core";

export function Draggable(props: { id: number }) {
  const { attributes, listeners, setNodeRef, transform } = useDraggable({
    id: props.id,
  });
  const style = transform
    ? {
        transform: `translate3d(${transform.x}px, ${transform.y}px, 0)`,
      }
    : undefined;

  return (
    <button ref={setNodeRef} style={style} {...listeners} {...attributes}>
      {props.id}
    </button>
  );
}

export function App() {
  return (
    <DndContext>
      <Draggable id={0} />
      <Draggable id={1} />
      <Draggable id={2} />
    </DndContext>
  );
}
```

Issue #1520, opened 2024-11-04
id issue


tanjiarui:
Hi there, I accidentally found the drag and drop operation failed when you setup the array index as `UniqueIdentifier`
for instance
```typescript jsx
const array = {'index': 0, 'item': 'a', 'index': 1, 'item': 'b'}
<SortableContext items={array.map(item => item['index'])}/>
```
you can never drag the first element in this case. i guess 0 recognized as `false`. is this situation deliberately designed?

Issue #1327, opened 2024-01-24
Cannot sort Elements with identifier={0}


LukasOchmann:
When using numbers as identifiers an element that recives the id={0} cannot be dragged/sorted.

[...]

Issue #1255, opened 2023-10-20
useSortable's id prop  


AshotHeb:
Sorting doesn't work properly when passing 0 ( number type ) to useSortable's id prop.

example 

```
    const id  = 0
    const { attributes, listeners, isDragging, isOver, node, setNodeRef } =
      useSortable({
        id
      })
```

PR #1542, opened 2024-11-23
Fix for when id is equal to 0

clauderic:
Fix a bug with draggable and sortable elements with an `id` equal to `0`.

Fixes https://github.com/clauderic/dnd-kit/issues/1532
Fixes https://github.com/clauderic/dnd-kit/issues/1520
Fixes https://github.com/clauderic/dnd-kit/issues/1327
Fixes https://github.com/clauderic/dnd-kit/issues/1255
Fixes https://github.com/clauderic/dnd-kit/issues/1150

### DIFF
--- a/.changeset/fix-zero-id.md
+++ b/.changeset/fix-zero-id.md
@@ -1,0 +1,6 @@
+---
+'@dnd-kit/core': patch
+'@dnd-ki/sortable': patch
+---
+
+Fix bug with draggable and sortable elements with an `id` equal to `0`.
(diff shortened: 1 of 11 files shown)


.changeset/fix-zero-id.md
```diff
@@ -0,0 +1,6 @@
+---
+'@dnd-kit/core': patch
+'@dnd-ki/sortable': patch
```
Grawl, line 3: typo

PR #1555, opened 2024-12-05
End keyboard drag operation when pressing tab

clauderic:
Resolves https://github.com/clauderic/dnd-kit/issues/857

This PR takes a different approach from https://github.com/clauderic/dnd-kit/pull/1087 and only de-activates the drag operation when the `Tab` key is pressed. This mostly accomplishes the same thing, but feels like a less risky change as there could be all sorts of factors that cause the draggable item to lose focus.

### DIFF
--- a/.changeset/end-on-tab.md
+++ b/.changeset/end-on-tab.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/core': patch
+---
+
+Added `Tab` to the list of default key codes that end a drag and drop operation. Can be customized by passing in a custom list of `keyCodes` to the KeyboardSensor options.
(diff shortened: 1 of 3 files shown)


jperals:
Hi, thank you for this improvement. Note that if you lose focus with the mouse, you still get an inconsistent state where the option is shown as being dragged but it isn't. Also, I can recover focus with arrow keys but then I can't move the items. That is why in #1087 drag is deactivated on a blur event (it doesn't matter what causes it), and this doesn't happen.

https://github.com/user-attachments/assets/1d4f6ecb-7585-4936-af0e-86df077322e5

joonass-visma:
@clauderic Is there any plans on fixing the mouse blur use-case that @jperals pointed out?

PR #1565, opened 2024-12-12
fix: rebind listeners on effect to address race condition

chrisvxd:
This fix rebinds the PointerSensor listeners in the effect callback, as the document may change between effects. The existing implementation yielded a race condition that meant that the listeners would sometimes not fire when rapidly dragging between frames.

Although this fix improves the behaviour, the issue sometimes still occurs and likely requires further investigation.

### DIFF
--- a/packages/dom/src/core/sensors/pointer/PointerSensor.ts
+++ b/packages/dom/src/core/sensors/pointer/PointerSensor.ts
@@ -83,13 +83,68 @@ export class PointerSensor extends Sensor<
 
         target.addEventListener('pointerdown', listener);
 
+        // Re-bind listeners if dragging, in case document changed during effect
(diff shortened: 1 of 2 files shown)


packages/dom/src/core/sensors/pointer/PointerSensor.ts
```diff
@@ -83,15 +83,70 @@ export class PointerSensor extends Sensor<
...
     });
 
     return unbind;
   }
 
+  protected bindListeners(
+    source: Draggable,
+    options: PointerSensorOptions = {}
+  ): () => void {
+    const target = source.handle ?? source.element;
```
clauderic, line 107: this was previously `event.target`

chrisvxd:
Closing in favour of #1566

PR #1566, opened 2024-12-12
fix: use fixed event listeners to address race conditions

chrisvxd:
Refactor the PointerSensor to use fixed event listeners, rather than dynamic ones, preventing race conditions that can occur if the document changed during drag, such as when dragging across frames.

To reproduce, rapidly drag items from the host to the iframe in the iframe stories.

### DIFF
--- a/packages/dom/src/core/sensors/pointer/PointerSensor.ts
+++ b/packages/dom/src/core/sensors/pointer/PointerSensor.ts
@@ -56,6 +56,8 @@ export class PointerSensor extends Sensor<
 
   protected initialCoordinates: Coordinates | undefined;
 
+  protected source: Draggable | undefined = undefined;
+
   #clearTimeout: CleanupFunction | undefined;
 
   constructor(
@@ -67,6 +69,14 @@ export class PointerSensor extends Sensor<
     this.handleCancel = this.handleCancel.bind(this);
     this.handlePointerUp = this.handlePointerUp.bind(this);
     this.handleKeyDown = this.handleKeyDown.bind(this);
+
+    effect(() => {
+      const unbindGlobal = this.bindGlobal(options ?? {});
+
+      return () => {
+        unbindGlobal();
+      };
+    });
   }
 
   public bind(source: Draggable, options = this.options) {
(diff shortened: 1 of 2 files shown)


packages/dom/src/core/sensors/pointer/PointerSensor.ts
```diff
@@ -92,6 +102,48 @@ export class PointerSensor extends Sensor<
...
+        documents.add(getDocument(draggable.element));
+      }
+    }
+
+    for (const droppable of this.manager.registry.droppables.value) {
+      if (droppable.element) {
+        documents.add(getDocument(droppable.element));
+      }
+    }
+
```
chrisvxd, line 119: Noting that this may be better elsewhere for perf reasons

packages/dom/src/core/sensors/pointer/PointerSensor.ts
```diff
@@ -145,48 +194,38 @@ export class PointerSensor extends Sensor<
...
+      return;
+    }
+
+    const ownerDocument =
+      this.source.element && getDocument(this.source.element);
+
+    // Event may have duplicated between documents if user is bubbling events
+    if (doc !== ownerDocument) {
+      return;
+    }
```
chrisvxd, line 221: Note, new safe-guard to account for an edge-case where the user bubbles the `pointermove` event to the parent document (as I am doing with Puck)

chrisvxd:
Moving back to draft as this change results in a bug, seen in https://github.com/measuredco/puck/issues/769. Unclear of resolution yet.

packages/dom/src/core/sensors/pointer/PointerSensor.ts
```diff
@@ -234,6 +273,10 @@ export class PointerSensor extends Sensor<
   }
 
   private handlePointerUp(event: PointerEvent) {
+    if (!this.source) {
+      return;
+    }
```
chrisvxd, line 278: Now we're using global event listeners, we need to check if the user is dragging before stopping propagation (see https://github.com/measuredco/puck/issues/769).

clauderic:
Can you provide more specifics about what race conditions this PR was meant to solve? I did notice in the iframe story that sometimes if you released the pointer in a different frame than the one where the drag operation originated, the drag operation would never end. In those cases, the `lostpointercapture` event gets fired so I added some logic to end the drag operation when that is fired here https://github.com/clauderic/dnd-kit/commit/0f1aa72a9414b5c08ebfae5715653ff4854f6646

chrisvxd:
It's hard to provide specifics, as I was only able to replicate on certain devices in certain environments (I even noticed differences when using an external display vs not an external display). Removing the dynamic behaviour addressed the issue, and generally seemed like a more reliable approach.

I'm happy to try and reproduce with the updated PointerSensor, but what's the reason for sticking with dynamic?

chrisvxd:
Unfortunately the latest dnd-kit (0.0.10-beta-20250212025300) still encounters this issue. Here's a video of the reproduction in Puck (wait for the 7th item to be added). It was previously possible to reproduce this in Storybook in the iframe story, although I haven't retested it this time.

https://github.com/user-attachments/assets/a440f899-a936-4ccb-8581-1b2df11d9b15

Using my custom PointerSensor still works.

clauderic:

Issue #338, opened 2021-06-22
DataTransfer


Eneris:
Hi. My use-case requires some dragging of data out of the site context / browser windows.
In other words, I need to use https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer

Is there a way to access Drag event to be able to set DataTransfer data?

Best,
Martin

clauderic:
There isn't. This is by design. Refer to https://github.com/clauderic/dnd-kit/discussions/306

Eneris:
Thanks for the quick response

myckhel:
I made this hack in place of DataTransfer
```tsx
const onDragStart = (e: DragStartEvent) => {
  if (e.active.data) {
    // @ts-ignore
    e.active.data.mode = 'form'
    // @ts-ignore
    e.active.data.type = 1
  }
}

function handleDragEnd(event) {
  console.log(event) // {active: {data: {current, type: 1, mode: 'form'}}}
}
  
<DndContext onDragEnd={handleDragEnd} onDragStart={onDragStart} />
```

PR #1584, opened 2025-01-07
Native file dragging plugin

yf-yang:
This PR aims at providing a way to drag and drop file from native fs to browser.

Currently it is marked as draft because:
- [ ] Some code structure needs to be discussed. Is `DraggableFileSymbol` at the right place? Is it OK to create a draggable for all the files in `dragOperation.ts`?
- [ ] I don't know why dragging file is not triggering collision observer. Needs more guidance.


packages/abstract/src/core/collision/observer.ts
```diff
@@ -90,7 +90,6 @@ export class CollisionObserver<
         continue;
       }
 
-      entry.shape;
```
yf-yang, line 93: Is it a typo?

packages/abstract/src/core/manager/dragOperation.ts
```diff
@@ -57,6 +53,8 @@ export type DragActions<
   V extends DragDropManager<T, U>,
 > = ReturnType<typeof DragOperationManager<T, U, V>>['actions'];
 
+export const DraggableFileSymbol = Symbol();
```
yf-yang, line 56: Is it OK to put it here?

yf-yang:
@clauderic Would you please give me some guidance of how collision observer works?

Specifically, I don't know which signal is triggering this effect:
https://github.com/clauderic/dnd-kit/blob/f3f05ba4d5e9dc24617c2072becde22eae1f5fe6/packages/abstract/src/core/collision/observer.ts#L24

With devtools, it seems that's from the position signal. However, when dragging files, I see the position signal does change, but the effect here is not fired. I don't know why.

packages/abstract/src/core/collision/observer.ts
```diff
@@ -67,9 +67,9 @@ export class CollisionObserver<
     collisionDetector?: CollisionDetector
   ) {
     const {registry, dragOperation} = this.manager;
-    const {source, shape, status} = dragOperation;
+    const {source, status} = dragOperation;
 
-    if (!status.initialized || !shape) {
+    if (!status.initialized) {
       return DEFAULT_VALUE;
```
yf-yang, line 73: It should be collision detection algorithm's responsibility to check shape's availability. File draggable does not have a shape and would be filtered by the original code implementation

packages/abstract/src/core/collision/observer.ts
```diff
@@ -90,7 +90,9 @@ export class CollisionObserver<
         continue;
       }
 
+      // Make sure effects will re-run when those properties change
       entry.shape;
+      dragOperation.position.current;
```
yf-yang, line 95: I don't know if it has a performance penalty, or why we don't simply remove the `untracked` call of `detectCollision`, so whatever signal used in `detectCollision` changes, the collision is recomputed.

packages/abstract/src/core/collision/observer.ts
```diff
@@ -67,9 +67,13 @@ export class CollisionObserver<
...
   ) {
     const {registry, dragOperation} = this.manager;
-    const {source, shape, status} = dragOperation;
+    const {source, status, position, shape} = dragOperation;
 
-    if (!status.initialized || !shape) {
+    // Make sure effects will re-run when those properties change
+    void position.current;
+    void shape;
+
```
yf-yang, line 75: I don't know if it has a performance penalty

yf-yang:
@clauderic Can we discuss here or how can we get connected? Generally speaking, to support native file drop, I've make sure they can be handled by dnd kit, but the native dom event needs to be passed to dnd-kit event handlers (Particularly, 'drag' event needs to be passed to 'dragend' event)

clauderic:
Hey can you help me better understand the use-case for this? I was considering introducing a native drag and drop plugin, for example, to allow drag and drop across windows, but I had not considered it for files.

yf-yang:
@clauderic It's simple: turning components to things like [react-dropzone](https://react-dropzone.js.org/#section-basic-example)

It's not the same with cross window drag and drop, as native file does not have a draggable source that could be captured by dnd-kit

Related #338 #306

PR #1590, opened 2025-01-23
React 19 (experimental)

chrisvxd:
Puck supports React 19, and uses it with the experimental branch of dnd-kit. It works fine, but dnd-kit's peer dependencies don't reflect it (which causes issues like https://github.com/measuredco/puck/issues/826).

This PR adds react 19 to the list of peer dependencies, and also upgrades the internally used version, addressing breaking changes to the type signature for useRef that now requires an argument ([see docs](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#useref-requires-argument)).

I didn't observe any issues, and it seems to build just fine.


packages/collision/package.json
```diff
@@ -33,7 +33,7 @@
     "@types/react-dom": "^18.0.4",
     "eslint": "^8.38.0",
     "@dnd-kit/eslint-config": "*",
-    "react": "^18.1.0",
+    "react": "^19.0.0",
```
clauderic, line 36: I don't think React is used in this package

```suggestion
```

clauderic (approved):

chrisvxd:
@clauderic pushed the requested change - believe this should be good to merge!

PR #1599, opened 2025-02-04
chore: port SnapCentertoCursor to experimental branch

khuezy:


khuezy:
@clauderic Does the `DragOverlay` respect the modifier? When I add the `DragOverlay`, it doesn't get centered on cursor.

packages/dom/src/modifiers/SnapCenterToCursor.ts
```diff
@@ -0,0 +1,28 @@
...
+    const offsetX = activatorCoordinates.x - left
+    const offsetY = activatorCoordinates.y - top
+
+    return {
+      ...transform,
+      x: transform.x + offsetX - width / 2,
+      y: transform.y + offsetY - height / 2,
+    }
+  }
+}
```
clauderic, line 28: I think you should move this to `packages/abstract/src/modifiers/snap.ts`. 

```suggestion
export class SnapCenterToCursor extends Modifier<
  DragDropManager<Draggable, Droppable>
> {
  apply({shape, position, transform}: DragOperation) {
    if (!position.initial || !shape) {
      return transform;
    }

    const {initial, current} = shape;
    const {left, top} = initial.boundingRectangle;
    const {height, width} = current.boundingRectangle;

    const offsetX = position.initial.x - left;
    const offsetY = position.initial.y - top;

    return {
      ...transform,
      x: transform.x + offsetX - width / 2,
      y: transform.y + offsetY - height / 2,
    };
  }
}
```

clauderic:
You should be able to add a story here: apps/stories/stories/react/Draggable/Modifiers/Modifiers.stories.tsx

Something like:

```ts
export const SnapToCursor: Story = {
  name: 'Snap to cursor',
  args: {
    modifiers: [SnapCenterToCursor],
  },
};
```

khuezy:
There is an issue w/ the typing when I apply the `modifier` to the `DragDropProvider`:

```
Type 'typeof SnapCenterToCursor' is not assignable to type 'ModifierConstructor<DragDropManager<Draggable<Data>, Droppable<Data>>> | ModifierDescriptor<DragDropManager<Draggable<Data>, Droppable<...>>>'.
  Type 'typeof SnapCenterToCursor' is not assignable to type 'ModifierConstructor<DragDropManager<Draggable<Data>, Droppable<Data>>>'.
    Types of parameters 'manager' and 'manager' are incompatible.
      Type 'DragDropManager<Draggable<Data>, Droppable<Data>>' is not assignable to type 'DragDropManager<Draggable<Data, DragDropManager<any, any>>, Droppable<Data, DragDropManager<any, any>>>'.
        Types of property 'collisionObserver' are incompatible.
          Type 'CollisionObserver<Draggable<Data>, Droppable<Data>, DragDropManager<Draggable<Data>, Droppable<Data>>>' is not assignable to type 'CollisionObserver<Draggable<Data, DragDropManager<any, any>>, Droppable<Data, DragDropManager<any, any>>, DragDropManager<Draggable<Data,
```

The code runs correctly but typescript complains.

yf-yang:
@khuezy There are two DragDropManager, one is from `@dnd-kit/abstract` and the other is from `@dnd-kit/dom`. That should be the cause

packages/dom/src/modifiers/index.ts
```diff
@@ -1,2 +1,3 @@
-export {RestrictToWindow} from './RestrictToWindow.ts';
-export {RestrictToElement} from './RestrictToElement.ts';
+export { RestrictToWindow } from './RestrictToWindow.ts';
+export { RestrictToElement } from './RestrictToElement.ts';
+export { SnapCenterToCursor } from './SnapCenterToCursor.ts';
```
clauderic, line 3:
```suggestion
```

packages/abstract/src/modifiers/snap.ts
```diff
@@ -8,15 +8,15 @@ import {
 } from '@dnd-kit/abstract';
 
 interface Options {
-  size: number | {x: number; y: number};
+  size: number | { x: number; y: number };
```
clauderic, line 11: I'm not sure why your editor isn't respecting the prettier settings, it should be:

```
{
  "bracketSpacing": false,
  "singleQuote": true,
  "semi": true,
  "trailingComma": "es5"
}
```

khuezy: Yea that's really strange. I see the prettier loading `.prettierrc` but it's not applying those settings.

PR #1639, opened 2025-03-25
feat(utilities): add closestLine and containingRectOrClosestLine algorithms

asterix696:
Two new collision detection strategies are introduced:

-  closestLine: Calculates the closest droppable container based on the distance between the pointer and the diagonal line of each container. This can provide more accurate results in scenarios where traditional bounding-box-based detection may fail.

- containingRectOrClosestLine: Returns the container if the pointer is inside its bounding rectangle. If not, it falls back to the closestLine strategy to find the closest container based on diagonal distance.

### DIFF
--- a/packages/core/src/utilities/algorithms/closestLine.ts
+++ b/packages/core/src/utilities/algorithms/closestLine.ts
@@ -1,0 +1,121 @@
+import type {Coordinates} from '../../types';
+import type {CollisionDescriptor, CollisionDetection} from './types';
+
+/**
+ * Returns the distance of the point from the given line.
+ */
+const getPointFromLineDistance = (
+  point: Coordinates,
+  [start, end]: [Coordinates, Coordinates]
+): number => {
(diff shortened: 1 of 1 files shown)


clauderic:
Can you elaborate on the benefits of this approach?

asterix696:
Certainly! These two strategies aim to address limitations in center or corner based collision detection, especially in complex or non-standard UI layouts enhancing UX in drag-and-drop where containers overlap or are tight.

They bring more value in these cases:

- closestLine: it works better in some layouts, especially with thin or custom-shaped droppable areas, where using center or corners can lead to unexpected or inaccurate detection. 

- containingRectOrClosestLine: given this strategy first checks if the pointer is inside a container, it prevents the misses when the pointer is technically inside a container but might be misclassified by proximity-based strategies. If the cursor is not in any container it has all the benefits of the closestLine algorithm.

clauderic:

PR #1655, opened 2025-04-17
fix: auto scroll on inverted element

blancham:
## Where
On a visually inverted element _(for example in a messaging app, your message list is inverted)_ 

## What
Approaching a dragged element to the edges of the inverted element scrolls in the opposite direction.


clauderic:
Code LGTM 👍  Would you mind adding a changeset?

blancham:
Sure, it's a first for me so tell me if it's right 🤓

clauderic (approved):

PR #1660, opened 2025-04-21
Add option `shouldActivate` on `KeyboardSensor`

GuillaumeSalles:
Currently `KeyboardSensor` does not activate if the event is coming from a descendent. `shouldActivate` lets us override this behavior.

### DIFF
--- a/packages/dom/src/core/sensors/keyboard/KeyboardSensor.ts
+++ b/packages/dom/src/core/sensors/keyboard/KeyboardSensor.ts
@@ -29,10 +29,19 @@ export type KeyboardCodes = {
 
 export interface KeyboardSensorOptions {
   keyboardCodes?: KeyboardCodes;
-  shouldActivate?(event: KeyboardEvent, source: Draggable): boolean;
+  shouldActivate?(args: {
+    event: KeyboardEvent;
+    source: Draggable;
+    manager: DragDropManager;
+  }): boolean;
 }
 
-const DEFAULT_SHOULD_ACTIVATE = (event: KeyboardEvent, source: Draggable) => {
+const DEFAULT_SHOULD_ACTIVATE = (args: {
+  event: KeyboardEvent;
+  source: Draggable;
+  manager: DragDropManager;
+}) => {
+  const {event, source} = args;
   const target = source.handle ?? source.element;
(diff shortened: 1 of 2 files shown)


packages/dom/src/core/sensors/keyboard/KeyboardSensor.ts
```diff
@@ -29,8 +29,14 @@ export type KeyboardCodes = {
 
 export interface KeyboardSensorOptions {
   keyboardCodes?: KeyboardCodes;
+  shouldActivate?(event: KeyboardEvent, source: Draggable): boolean;
```
clauderic, line 32:
```suggestion
  shouldActivate?(args: {event: KeyboardEvent, source: Draggable, manager: DragDropManager}): boolean;
```

packages/dom/src/core/sensors/keyboard/KeyboardSensor.ts
```diff
@@ -99,10 +105,9 @@ export class KeyboardSensor extends Sensor<
...
-    if (
-      (!source.handle && source.element && event.target === source.element) ||
-      (source.handle && event.target === source.handle)
-    ) {
+    const shouldActivate = options?.shouldActivate ?? DEFAULT_SHOULD_ACTIVATE;
+
+    if (shouldActivate(event, source)) {
       const {keyboardCodes = DEFAULT_KEYBOARD_CODES} = options ?? {};
 
       if (!keyboardCodes.start.includes(event.code)) {
```
clauderic, line 113: There's a bit of tension between `shouldActivate` and `keyboardCodes`, not sure whether we should leave this here or move it into the default `shouldActivate`

packages/dom/src/core/sensors/keyboard/KeyboardSensor.ts
GuillaumeSalles: Done

packages/dom/src/core/sensors/keyboard/KeyboardSensor.ts
GuillaumeSalles: Updated to reflect our private discussion

clauderic (approved):

PR #1668, opened 2025-04-28
feat: `@dnd-kit/vue`

hanneskuettner:
### Vue wrapper for `dnd-kit`

A somewhat small wrapper mapping the dnd-kit core to the Vue reactivity system. 

_WIP_

Currently doing a self review of all the changes, but I wanted to put this out there, for other people to start having a look.

#### Notes

This wrapper tries to maintain the same API as the React hooks provided in `@dnd-kit/react`, notable differences being:

* `hooks` are renamed to `composables` to fit the Vue nomenclature
* The `useDraggable`/`useDroppable`/`useSortable` composables only accept an `Element` `ref`, but don't return a bindable `ref` for the element, handle, etc. This is a lot more common in the Vue world, to the best of my knowledge, e.g. as seen in this popular composables library: https://vueuse.org/core/useElementSize/

This allows for the following code, to make an element draggable
```vue
<script lang="ts" setup>
  import { useTemplateRef } from 'vue';
  import { useDraggable } from '@dnd-kit/vue';

  const elRef = useTemplateRef('el');
  useDraggable({
    id: 'draggable',
    element: elRef
  });
</script>

<template>
  <button ref="el">
    Draggable
  </button>
</template>
```


packages/vue/src/sortable/useSortable.ts
```diff
@@ -0,0 +1,113 @@
...
+        element: unrefElement(input.element),
+        handle: unrefElement(input.handle),
+        target: unrefElement(input.target),
+      },
+      manager
+    );
+  });
+  const trackedSortable = useDeepSignal(sortable);
+
+  watchEffect(() => {
```
clauderic, line 51: Should / can these use `batch`?

packages/vue/src/sortable/useSortable.ts
```diff
@@ -0,0 +1,113 @@
...
+  element?: MaybeRefOrGetter<MaybeElement>;
+  source?: MaybeRefOrGetter<MaybeElement>;
+  target?: MaybeRefOrGetter<MaybeElement>;
+}
+
+export function useSortable<T extends Data = Data>(input: UseSortableInput<T>) {
+  const transition = computed(() => ({
+    ...defaultSortableTransition,
+    ...toValue(input.transition),
+  }));
```
clauderic, line 32: Can we delegate this responsibility to `@dnd-kit/dom`? I don't remember why I did this in `@dnd-kit/react` in the first place and not in `@dnd-kit/dom` instead

packages/vue/package.json
```diff
@@ -0,0 +1,78 @@
+{
+  "name": "@dnd-kit/vue",
+  "version": "0.1.0",
```
clauderic, line 3: Leaving a note here for us to make sure we sync this with whatever the latest versions are of other packages before merging

packages/vue/src/core/context/DragDropProvider.ts
```diff
@@ -0,0 +1,137 @@
...
+        manager.value = _manager;
+
+        onWatcherCleanup(() => cleanupFns.forEach((fn) => fn()));
+      },
+      {
+        immediate: true,
+      }
+    );
+
+    watchEffect(() => {
```
clauderic, line 121: Can we use `batch` here also?

packages/vue/src/core/context/renderer.ts
```diff
@@ -0,0 +1,48 @@
...
+  );
+
+  const renderer = {
+    get rendering() {
+      return rendering.value;
+    },
+  };
+
+  function trackRendering(callback: () => void) {
+    if (rendering.value === Promise.resolve()) {
```
clauderic, line 32: Can this ever be true?
Screenshot 2025-04-28 at 7 17 02 PM

packages/vue/src/core/context/renderer.ts
```diff
@@ -0,0 +1,48 @@
...
+    {
+      flush: 'post',
+    }
+  );
+
+  const renderer = {
+    get rendering() {
+      return rendering.value;
+    },
+  };
```
clauderic, line 29: Does this need to be memoized at all? Not super familiar with vue best practices

clauderic (commented):
Heroic effort 🙌  Left a few initial questions as I was reading through the code, I haven't used vue much so I definitely learned a lot!

packages/vue/src/core/context/renderer.ts
denishsharma: In Vue, reactivity is opt-in unlike in react it's opt-out, that's why react has `useMemo`, `useCallback`, etc. So script will run only once, if we want to make it reactive, then we will have to opt-in using `ref`, `computed`, `watch`, `watchEffect` etc.

So no memoization is needed.

3dfactor:
Is this moving forward? What needs to be done for this to get merged/supported?

hanneskuettner:
Closed in favor of #1880

PR #1676, opened 2025-05-03
Fix typo: rename `sevent` to `event` in onDragEnd handler

Matteo002Turcato:
Corrected a typographical error in the onDragEnd handler where the variable `event` was mistakenly written as `sevent`.


clauderic:
Thanks!

PR #1690, opened 2025-05-07
Parse transform fix

GuillaumeSalles:
Defensive check in case `getComputedStyle().transform` is `null`. Could be the case on < Chrome 103


GuillaumeSalles:
Not the right fix.

Issue #1738, opened 2025-06-12
[experimental] Drag area not restricted to handle when using separate hooks


valtism:
I'm trying to implement a table with row reordering using dnd-kit, like the example given here https://tanstack.com/table/latest/docs/framework/react/examples/row-dnd

Because of the way TanStack Table works, the component where I need to attach the `ref` and the component to use as the drag handle are in different places where I cannot easily pass around a ref from a common ancestor.

In the example given, they use two separate hooks with the same id, and that seems to give the wanted behaviour of the row only becoming draggable from the handle.

It seems that when I do it using the new version, it only makes the row not draggable when the `handleRef` is applied from the same hook that is supplying the `ref`.

I'm not sure what the intended behaviour is here, but it would be helpful to be able to get a handleRef from a separate hook and have interaction connected automatically through the DragDropProvider.

I'm sure I can work something with custom context, but just want to know if this is a bug or not.

clauderic:
I'm surprised that guidance ever worked to be quite honest. Try this out https://github.com/clauderic/dnd-kit/pull/1740

valtism:
That works perfectly, thank you! I was able to make it work with a custom context, but this is more ergonomic, even though it does expand the API surface

clauderic:
Looks like this was resolved by @clauderic — PR #1740 adds support for this use case and the reporter confirmed it works. Closing this issue. Thanks for the report and for testing!

PR #1740, opened 2025-06-12
[Experimental] Introduce `useDragHandle` hook

clauderic:
This PR introduces a `useDragHandle` hook that allows React consumers to set a drag handle in a different location than where `useDraggable` or `useSortable` is used.

Resolves https://github.com/clauderic/dnd-kit/issues/1738

### Example usage — two separate components

Below the draggable is declared in a row component, while the dedicated handle lives in a different component.

```tsx
import {DragDropProvider} from '@dnd-kit/react';
import {useSortable} from '@dnd-kit/react/sortable';
import {useDragHandle} from '@dnd-kit/react';

import {
  ColumnDef,
  Row,
  flexRender,
  useReactTable,
  getCoreRowModel,
} from '@tanstack/react-table';

// ─────────────────────────────────────────────────────
// Cell renderer — only cares about being the drag handle
// ─────────────────────────────────────────────────────
function DragCell({rowId}: {rowId: string}) {
  const handleRef = useDragHandle({id: rowId});   // same id as the row
  return (
    <button ref={handleRef} aria-label="Drag row" style={{cursor: 'grab'}}>
      ⠿
    </button>
  );
}

// ─────────────────────────────────────────────────────
// Draggable row — owns the sortable behaviour
// ─────────────────────────────────────────────────────
function DraggableRow({row}: {row: Row<Person>}) {
  const {ref} = useSortable({id: row.id, index: row.index});        // row itself is sortable
  return (
    <tr ref={ref}>
      {row.getVisibleCells().map((cell) => (
        <td key={cell.id}>{flexRender(cell.column.columnDef.cell, cell.getContext())}</td>
      ))}
    </tr>
  );
}

// ─────────────────────────────────────────────────────
// Table setup (irrelevant plumbing omitted for clarity)
// ─────────────────────────────────────────────────────
function App() {
  const columns = React.useMemo<ColumnDef<Person>[]>(() => [
    {                                             // dedicated “move” column
      id: 'drag',
      header: '↕︎',
      size: 40,
      cell: ({row}) => <DragCell rowId={row.id} />, // ← handle rendered here
    },
    // …other columns…
  ], []);

  const table = useReactTable({
    data, columns, getCoreRowModel: getCoreRowModel(), getRowId: r => r.id,
  });

  return (
    <DragDropProvider>
      <table>
        <tbody>
          {table.getRowModel().rows.map((row) => (
            <DraggableRow key={row.id} row={row} />
          ))}
        </tbody>
      </table>
    </DragDropProvider>
  );
}
```

The key takeaway: `useDragHandle` does **not** need to live in the same component where `useDraggable`/`useSortable` is invoked—perfect for table rows, list items with nested actions, or any layout where the “grab-area” is rendered deeper in the tree.

### DIFF
--- a/packages/react/src/core/draggable/useDragHandle.ts
+++ b/packages/react/src/core/draggable/useDragHandle.ts
@@ -1,0 +1,36 @@
+import {useRef} from 'react';
+import type {UniqueIdentifier} from '@dnd-kit/abstract';
+import {useComputed, useOnElementChange} from '@dnd-kit/react/hooks';
+import type {RefOrValue} from '@dnd-kit/react/utilities';
+
+import {useDragDropManager} from '../hooks/useDragDropManager.ts';
+
+export interface UseDragHandleInput {
+  id: UniqueIdentifier;
+  element?: RefOrValue<Element>;
+}
+
+/**
+ * A hook that allows you to set the drag handle for a draggable element.
+ *
+ * @param id - The unique identifier of the draggable element.
+ * @param element - The ref or value of the drag handle element if you already have a reference to it.
+ * @returns A ref that you can attach to the drag handle element
+ */
+export function useDragHandle({id, element}: UseDragHandleInput) {
+  const manager = useDragDropManager();
+  const handleRef = useRef<Element | null>(null);
+  const draggableInstance = useComputed(
+    () => manager?.registry.draggables.get(id),
+    [id]
+  ).value;
+  const value = handleRef.current ?? element;
+
+  useOnElementChange(value, (handle) => {
+    if (!draggableInstance) return;
+
+    draggableInstance.handle = handle;
+  });
+
+  return handleRef;
+}
--- a/packages/react/src/core/index.ts
+++ b/packages/react/src/core/index.ts
@@ -9,6 +9,10 @@ export {
   useDraggable,
   type UseDraggableInput,
 } from './draggable/useDraggable.ts';
+export {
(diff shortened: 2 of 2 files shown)


packages/react/src/core/draggable/useDragHandle.ts
```diff
@@ -0,0 +1,36 @@
...
+  const draggableInstance = useComputed(
+    () => manager?.registry.draggables.get(id),
+    [id]
+  ).value;
+  const value = handleRef.current ?? element;
+
+  useOnElementChange(value, (handle) => {
+    if (!draggableInstance) return;
+
+    draggableInstance.handle = handle;
```
valtism, line 32: This line does get flagged by the react compiler lint rule, but I can see this is using a ref inside a custom signals implementation so no idea if that the compiler is even a viable target for the library.

```
Mutating a value returned from a function whose return value should not be mutated (eslint react-compiler/react-compiler)
```

packages/react/src/core/draggable/useDragHandle.ts
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  return handleRef;
```
valtism, line 35: The return type here is `RefObject<Element | null>`, so you'll probably want to match that to the existing handleRef type of `(element: Element | null) => void` or at least allow a generic for the element type so it doesn't complain when being applied to something like a `<button>`

PR #1749, opened 2025-06-24
[experimental] Add documentation for getting reordered state using `move` in `useSortable` doc

sazid:


rrmahalo:
I don't recommend this being added in with the way `move` currently functions. There is some very confusing behavior in there, largely from the following:

```
const findIndex = (item: Items[0], id: UniqueIdentifier) =>
  item === id || (typeof item === 'object' && 'id' in item && item.id === id);
```

This completely ignores the id that's passed into the useSortable hook, instead relying on the existence of an id field within the item, which in turn breaks the ability to sort the controlled list when using any sort of computed id or a data structure with an unconventional natural id. There's also the issue of the optimistic update obfuscating the successful drag operation, since both source and target fields end up with the same id. I think the only current workaround is to use the initialIndex and index fields present on the sortable field within both the operation source and target. These aren't present on the provided types but it works.

songkeys:
or, convert it to sortable id before `move` then convert it back after:

```ts
		const sortableIds = ids.map(id => makeSortableId(id))
		const nextSortableIds = move(sortableIds, event)
		const nextIds = nextSortableIds.map(id => unwrapSortableId(id))
		persist(nextIds)
```

```ts
		useSortable({ id: makeSortableId(id) })
```

PR #1775, opened 2025-08-06
Add nonce option to Cursor, PreventSelection plugins for style elements

fpronto:
I just added the suppor for nonce in this two plugins, as they are style elements.


clauderic (approved):
Thanks for taking this on, changes look good to me. I may look into finding a more centralized way to do this in the future but for now this looks good to me 👍

luca-workera:
I wanted to say thank you for this improvement: it could be very important in an application that runs a strict CSP config.
I am writing to suggest to add it to the documentation, I could not find it there but this is valuable info 🙏

Issue #1794, opened 2025-09-08
[feature request] Disable autoScroll in the horizontal and vertical directions separately


ilovejlj:
The situation is, I have a table which can scroll both horizontally and vertically. Then I want to add the ability to drag and sort the rows.

In this situation, I wish the container cab scroll vertically but not horizontally.

The 'autoScroll' option of DndContext can disable both horizontal and vertical scroll of parent element together when set to false, but I wish I can only disable vertical scroll.

<img width="1719" height="303" alt="Image" src="https://github.com/user-attachments/assets/f2804a66-681f-4715-b30f-18bda4f45626" />

iPommes:
Hi there,
technically there is already a possibility to achieve this.

My solution to this problem is:

`<DndContext
      autoScroll={{
        threshold: {
          x: 0,
          y: 0.2,
        },
      }}
    >`

You can simply set the threshold from the autoScroll object on the DndContext to 0. This will disable the autoscroll in this direction.

For the other axis you can use your own value or use the default one (which is 0.2 on both axis).

I hope this helps, since I had the same problem and had to research a bit in the DnD-Kit codebase.

ilovejlj:
> Hi there, technically there is already a possibility to achieve this.
> 
> My solution to this problem is:
> 
> `<DndContext autoScroll={{ threshold: { x: 0, y: 0.2, }, }} >`
> 
> You can simply set the threshold from the autoScroll object on the DndContext to 0. This will disable the autoscroll in this direction.
> 
> For the other axis you can use your own value or use the default one (which is 0.2 on both axis).
> 
> I hope this helps, since I had the same problem and had to research a bit in the DnD-Kit codebase.

@iPommes Thanks for the advise. I tried to extends the MouseSensor and that seems to work:

``` ts
export class VerticalMouseSensor extends MouseSensor {
  constructor(props: MouseSensorProps) {
    const originOnMove = props.onMove;
    const originOnStart= props.onStart;
    super({
      ...props,
      onStart: (coordinates: Coordinates) => {
        originOnStart({
          ...coordinates,
          x: 0,
        })
      },
      onMove: (coordinates: Coordinates) => {
        originOnMove({
          ...coordinates,
          x: 0,
        })
      },
    });
  }
}
```

`{ threshold: { x: 0, y: 0.2, }, }` also works but that finally triggers `scrollTo(Infinity, positionY)` which I am not sure if it works on all browsers, because it's not described in standard  about using Infinity

[...]

PR #1795, opened 2025-09-08
prevent vertical or horizontal scrolling individually

cwj0417:
hi, thank you for your great work on dnd-kit.

there is a scenario where we need to drag a table with many columns vertically, and we hope that this scenario will prevent horizontal scrolling during dragging. (#1794)

we can already prevent vertical or horizontal scrolling individually by setting threshold.x or threshold.y to 0, but we call `scrollBy(Infinity, {number})` actually, I set speed and direction to 0 when threshold set to 0.


clauderic:

Issue #1835, opened 2025-11-12
DragStartEvent not exposed in experimental versions


mattrunyon:
Docs show a `DragStartEvent` like the core version https://next.dndkit.com/react/components/drag-drop-provider#param-on-drag-start

However, these types are not exposed. They are indirectly exposed via a (seemingly very complex) generic type `DragDropEvents['dragStart']`

I think the individual event types should be exposed as they were before and/or the event handler function types that include both params.

clauderic:
Feel free to submit a PR, would gladly accept it. Will get to it when I can otherwise.

PR #1837, opened 2025-11-16
Expose the DragDropEvent types for people using TypeScript

bagedevimo:
These are currently exposed via the DragDropEvent which are pretty brutal to use a in a case, these are just tiny type aliases to make it a little easier to use.

Closes issue #1835


bagedevimo:
Should I add a changeset to this PR? It's not clear if you'd like one for adding new types

PR #1840, opened 2025-11-23
fix: import path for RestrictToWindow in docs

njacob1001:
the import path in the documentation is wrong, the correct should be:

```js
import { RestrictToWindow } from '@dnd-kit/dom/modifiers';
```

instead of
```js
import { RestrictToWindow } from '@dnd-kit/react';
```

### DIFF
--- a/apps/docs/react/components/drag-drop-provider.mdx
+++ b/apps/docs/react/components/drag-drop-provider.mdx
@@ -89,8 +89,8 @@ import {
   DragDropProvider,
   PointerSensor,
   KeyboardSensor,
-  RestrictToWindow,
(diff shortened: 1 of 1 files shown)


clauderic (approved):
## Summary
Fixes an incorrect import path in the docs — `RestrictToWindow` is not exported from `@dnd-kit/react`, only from `@dnd-kit/dom/modifiers`. This is a straightforward docs correction.

## Feedback

### Changeset
- Status: not present — correct, this is a docs-only change.

## Overall
Small, correct fix. Thanks for catching this, @njacob1001!

---
[claude-review]

PR #1854, opened 2025-12-23
[next] feedback style injection in shadow dom (fix #1765)

### LINKED ISSUE #1765: [next] Shadow dom support?
The current positioning of the "cloned pseudo-dragging element" (how do you name it?) is problematic inside a shadow dom.

Preview:

Image

CodeSandBox:
https://codesandbox.io/p/sandbox/keen-bell-jm6lkv

du33169:
Fix #1765.

- Updated CSS_RULES with additional shadow host selector: 
  `:root [${ATTRIBUTE}]` -> `:is(:root,:host) [${ATTRIBUTE}]`. 
- Added two utility functions:
  - `isShadowRoot(target: EventTarget | null): target is ShadowRoot`
  -  `getRoot(target): Document | ShadowRoot`
- Updated `#injectStyles()` to support injecting styles to shadow root.

Now the feedback element moves properly during dragging.

Some other `getDocument` invocations could also be updated to make the library less intrusive, for example, by restricting the listener target or auto-scrolling within the shadow tree, instead of document-wide. But currently these are left unchanged as everything seems to work just fine.

---

### DIFF
--- a/packages/dom/src/core/plugins/feedback/Feedback.ts
+++ b/packages/dom/src/core/plugins/feedback/Feedback.ts
@@ -2,6 +2,7 @@ import {
   effects,
   reactive,
   untracked,
+  derived,
   type CleanupFunction,
 } from '@dnd-kit/state';
 import {configurator, Plugin} from '@dnd-kit/abstract';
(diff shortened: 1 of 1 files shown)


packages/dom/src/core/plugins/feedback/Feedback.ts
```diff
@@ -677,33 +678,46 @@ export class Feedback extends Plugin<DragDropManager, FeedbackOptions> {
...
+        return root.head;
+      }
+      function injectStyleToRoot(root: Document | ShadowRoot, style: HTMLStyleElement) {
+        if (isShadowRoot(root)) {
+          root.prepend(style);
+        }
+        else if (root instanceof Document) {
+          root.head.prepend(style);
+        }
+      }
```
clauderic: Could we adopt the stylesheet into the shadow root instead? 

https://developer.mozilla.org/en-US/docs/Web/API/Document/adoptedStyleSheets

du33169:
> This is looking good overall, though I'm curious how this behaves if you initiate a drag and drop operation in a Shadow Root but then move outside of it or into another root

I created a simple story to test it. When dragging an element into a shadow DOM for the first time, the dropping animation breaks indeed. This happens because the styles are only injected during `state.initializing` (when dragging starts), but at that time the drop target isn’t determined yet, so the styles never get injected into the shadow DOM. Does the `state.initializing` condition matter? Removing it fixes the animation but I assume it was added for some reason.

This problem should also apply to iframe, but it might be covered up by the AutoFrame component (which automatically sync the styles to the iframe) used in iframe examples.

packages/dom/src/core/plugins/feedback/Feedback.ts
clauderic: We can use adoptedStylesheets for both 👍

PR #1873, opened 2026-02-03
Fix `sortableKeyboardCoordinates` offset calculation bug

reteps:
## Summary

- The faulty conditional `offset.x && offset.y ? rectCoordinates : subtract(rectCoordinates, offset)` skipped offset subtraction when both x and y offsets were non-zero

## Test plan

- [x] Added unit tests covering offset application in various scenarios
- [x] Added regression test that fails with old code, passes with fix

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/large-readers-smell.md
+++ b/.changeset/large-readers-smell.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/sortable': patch
+---
+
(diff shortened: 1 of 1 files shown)


clauderic:

PR #1946, opened 2026-03-02
fix(plugins): recalculate `AutoScroller` options in the effect to avoid stale data

### LINKED ISSUE #1945: `AutoScroller` doesn't respect updated options
`AutoScroller` calculates its options only once when the instance is created:

https://github.com/clauderic/dnd-kit/blob/f3f09cc00f9488590506f5532d5c13513938ed80/packages/dom/src/core/plugins/scrolling/AutoScroller.ts#L39

That means that even if options are updated here:

https://github.com/clauderic/dnd-kit/blob/f3f09cc00f9488590506f5532d5c13513938ed80/packages/abstract/src/core/plugins/registry.ts#L61

they're still going to be the same options that were initially passed to the constructor.

**Reproduction link:** https://codesandbox.io/p/sandbox/elated-colden-f2vkc6?file=/App.js

**Steps to reproduce:**
1. Try dragging an item from the second column to the first one.
2. Notice that the list is auto-scrollable in both directions (items get pushed even horizontally).
3. Click the 'Update Options' button.
4. Observe that nothing changes.

**Additional info:** Reproducible on the latest beta version.

**Solution:** Move `scrollOptions` into the effect itself and access them via `this.options` instead of using `options` from the constructor.

mattersj:
fixes #1945

### DIFF
--- a/.changeset/whole-cloths-warn.md
+++ b/.changeset/whole-cloths-warn.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/dom': patch
+---
+
+recalculate AutoScroller options in the effect to avoid stale data


clauderic:
Thank you!

PR #1973, opened 2026-03-28
fix(accessibility): ensure screen reader announcements update consistently

### LINKED ISSUE #1952: dnd-kit-announcement stops updating when item is moved
When dragging a sortable item up or down in a vertical list, the live-region element:

**id="dnd-kit-announcement-4"**

stops updating its content.

Because of this, screen readers no longer announce position changes or drag state updates after some movements.

This creates an accessibility regression — the drag interaction becomes silent for assistive technology users.

https://main--5fc05e08a4a65d0021ae0bf2.chromatic.com/?path=/story/react-sortable-vertical-list--basic-setup

cyphercodes:
Fixes #1952 - The useAnnouncement hook now clears the announcement before setting the new value, using requestAnimationFrame to ensure screen readers detect the change.

### DIFF
--- a/.changeset/stale-owls-search.md
+++ b/.changeset/stale-owls-search.md
@@ -1,0 +1,5 @@
+---
+"@dnd-kit/accessibility": patch
+---
+
+Fix screen reader announcements so repeated messages are announced consistently.
--- a/packages/accessibility/src/hooks/useAnnouncement.ts
+++ b/packages/accessibility/src/hooks/useAnnouncement.ts
@@ -4,7 +4,13 @@ export function useAnnouncement() {
   const [announcement, setAnnouncement] = useState('');
   const announce = useCallback((value: string | undefined) => {
     if (value != null) {
-      setAnnouncement(value);
+      // Clear the announcement first to ensure screen readers detect the change
+      // even when the same text is announced consecutively
+      setAnnouncement('');
+      // Use requestAnimationFrame to ensure the clear renders before setting new value
+      requestAnimationFrame(() => {
+        setAnnouncement(value);
+      });
     }
   }, []);
 


clauderic (changes requested):
## Summary

This PR fixes an issue where screen readers don't re-announce identical text content. It uses a clear-then-set pattern via `requestAnimationFrame` to force the screen reader to detect the change.

## Feedback

### Blocking issues

- **Missing changeset**: `packages/accessibility/src/hooks/useAnnouncement.ts` is a source file in a published package, so a changeset entry is required. Please add a `.changeset/*.md` file listing `@dnd-kit/accessibility` with a `patch` bump.

- **Base branch**: This PR targets `master` (the legacy branch), not `main` (the active development branch). Please double-check which branch you intended to target. If this is intentional for the legacy package, that's fine — just worth confirming.

### Suggestions

- `packages/accessibility/src/hooks/useAnnouncement.ts` — The `requestAnimationFrame` callback isn't cancelled if the component unmounts before the frame fires. While React 18 makes this a no-op (no warning), a cleaner pattern would be to return a cancel handle. For example:

  ```ts
  const announce = useCallback((value: string | undefined) => {
    if (value != null) {
      setAnnouncement('');
      const rafId = requestAnimationFrame(() => {
        setAnnouncement(value);
      });
      return () => cancelAnimationFrame(rafId);
    }
  }, []);
  ```

  This is a suggestion, not a blocker — the current approach will work correctly in React 18+.

### Changeset

- **Status**: missing. A changeset is required for changes to `packages/*/src/`.

## Overall

Great fix for a subtle screen reader issue — the clear-then-rAF pattern is the right approach here. Thanks for contributing! Once the changeset is added, this looks good to go.

---
[claude-review]

cyphercodes:
Thanks for the review. I pushed `614e8c2b` with the required changeset:

- Added `.changeset/stale-owls-search.md` with a patch bump for `@dnd-kit/accessibility`.
- Verified it with `npx --yes @changesets/cli@2.16.0 status --since=HEAD~1`; it reports `@dnd-kit/accessibility` as a patch bump.

I left the PR targeting `master`: this branch's changeset config uses `baseBranch: "master"`, and the touched legacy package path (`packages/accessibility/src/hooks/useAnnouncement.ts`) does not exist on `main`'s current package layout. If you prefer a separate `main`-branch implementation, I can handle that separately.

BohdanZinkevych:
Hi! Any updates on this review?

PR #1988, opened 2026-04-02
fix: set operation.shape before computing initial transform in Feedback plugin

### LINKED ISSUE #1986: Operation.shape is null on first modifier.apply() call
Hello there, 

I am using dnd-kit/react@0.3.2  and trying to create my custom modifier, that will snap topleft corner of the dragged object to the mouse cursor. Here is what I have so far: 

```
import {
    DragDropManager,
    Modifier,
    PluginOptions,
} from "@dnd-kit/abstract";
import { Coordinates } from '@dnd-kit/geometry';

export class snapLeftTopToCursor extends Modifier {
    constructor(manager: DragDropManager<any, any>, options?: PluginOptions) {
        super(manager, options)
    }

    public apply(operation: any): Coordinates {
        const { transform, activatorEvent, shape } = operation

        const rect = shape?.initial
        if (this.disabled || !rect) return transform;

        const offsetX = activatorEvent?.clientX - rect.left;
        const offsetY = activatorEvent?.clientY - rect.top;

        return {
            ...transform,
            x: transform.x + offsetX,
            y: transform.y + offsetY,
        }
    }
}
```
The problem is that when the `apply()` is called for the first time, `shape` is `null`, and so I can not calculate new tranform coordinates. 

This is just a migration of a modifier that was created for the legacy dnd-kit, and it works as expected: 

```
import { Modifier } from "@dnd-kit/core";
import { getEventCoordinates } from '@dnd-kit/utilities';

export const snapLeftTopToCursor: Modifier = ({ transform, activatorEvent, draggingNodeRect }) => {
    if (draggingNodeRect && activatorEvent) {
        const activatorCoordinates = getEventCoordinates(activatorEvent);

        if (!activatorCoordinates) {
            return transform;
        }

        const offsetX = activatorCoordinates.x - draggingNodeRect.left;
        const offsetY = activatorCoordinates.y - draggingNodeRect.top;

        return {
            ...transform,
            x: transform.x + offsetX,
            y: transform.y + offsetY,
        };
    }

    return transform;
}

```

clauderic:
Fixes #1986

## Root cause

In `Feedback.ts`, the initial transform was computed (line 313) *before* the feedback element's shape was measured and set on `dragOperation` (lines 391–392). This meant that on the very first call to `modifier.apply()`, `operation.shape` was always `null`.

The ordering was:
1. `feedbackElement.setAttribute(ATTRIBUTE, 'true')` — gives element `position: fixed`
2. `dragOperation.transform` read → modifiers called with `shape = null` ❌
3. `styles.set(...)` — positions the feedback element correctly
4. placeholder setup, popover promotion, observers
5. `const initialShape = new DOMRectangle(feedbackElement)` → `dragOperation.shape = initialShape`

## Fix

Reorder so styles are applied first, *then* shape is measured and set, *then* the transform is computed:

1. `styles.set(...)` with `initialTranslate` (no drag-transform offset yet)
2. placeholder setup, popover promotion, observers (unchanged)
3. `const initialShape = new DOMRectangle(feedbackElement)` → `dragOperation.shape = initialShape`
4. `dragOperation.transform` read → modifiers now have `shape.initial` ✅
5. `styles.set({ translate })` — update translate with the modifier-adjusted value

On the first frame `position.delta` is `{x:0,y:0}`, so for modifiers that don't use shape the second `styles.set` is a no-op. For shape-dependent modifiers (snap-to-cursor etc.) it correctly applies the adjusted translation.

This restores the v1 behaviour where `draggingNodeRect` was always populated when a modifier ran.

## Test plan
- Create a snap-to-cursor modifier that reads `operation.shape?.initial` — verify the rect is non-null on the first frame
- Verify existing drag behaviour (no modifier, SnapModifier, BoundingRectangle modifier) is unchanged
- Test with and without `DragOverlay`

---
*Automated fix by Claude Code*

### DIFF
--- a/.changeset/fix-shape-null-first-modifier-call.md
+++ b/.changeset/fix-shape-null-first-modifier-call.md
@@ -1,0 +1,5 @@
+---
+"@dnd-kit/dom": patch
+---
+
+Fixed `operation.shape` being `null` on the first `modifier.apply()` call. The `Feedback` plugin now measures the feedback element's shape and sets it on the drag operation *before* computing the initial transform, so modifiers that depend on `shape.initial` (e.g. snap-to-cursor) receive the correct bounding rect on the first frame. This restores the v1 behaviour where `draggingNodeRect` was always available in modifier functions.
--- a/apps/stories-shared/tests/modifiers.tests.ts
+++ b/apps/stories-shared/tests/modifiers.tests.ts
@@ -1,0 +1,159 @@
+import {test, expect} from './fixtures.ts';
+
+interface ModifierStories {
+  verticalAxis: string;
+  horizontalAxis: string;
+  restrictToWindow: string;
+  snapToGrid: string;
+}
+
+export function modifierTests(stories: ModifierStories) {
+  test.describe('Modifiers', () => {
+    test.describe('RestrictToVerticalAxis', () => {
+      test('restricts drag movement to the vertical axis only', async ({
+        dnd,
+      }) => {
+        await dnd.goto(stories.verticalAxis);
+        await dnd.disableTransitions();
+
+        const button = dnd.buttons.first();
+        await expect(button).toBeVisible();
+
+        const box = await button.boundingBox();
+        if (!box) throw new Error('Could not get bounding box');
+
+        const startX = box.x + box.width / 2;
+        const startY = box.y + box.height / 2;
+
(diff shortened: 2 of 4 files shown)


clauderic (commented):
❌ Broken animation on multiple lists keyboard sorting

PR #2017, opened 2026-04-16
fix: (legacy version) fix utils used for calculating scroll offsets inside of iframes

istateside:
This PR is for the legacy version of the library - i'm not sure if you're still accepting patches for the @dnd-kit/core and @dnd-kit/utilities libraries, but these were blocking bugs for me

I noticed that scroll offsets are not calculated correctly when the draggable nodes are in a different frame than the execution context.

`getScrollableElement` was returning `window` directly instead of fetching the correct window object off of the `element` argument
and `getWindow` was not properly handling the case where the given `target` element was a Document object from inside of an iframe, which can be the case if the iframe document has no scrollable element beyond the `html` node

### DIFF
--- a/packages/utilities/src/execution-context/getWindow.ts
+++ b/packages/utilities/src/execution-context/getWindow.ts
@@ -14,5 +14,9 @@ export function getWindow(target: Event['target']): typeof window {
     return window;
   }
 
-  return target.ownerDocument?.defaultView ?? window;
+  let ownerDoc = target.ownerDocument;
+  if (!ownerDoc) {
+    ownerDoc = 'documentElement' in target ? target : null;
+  }
+  return ownerDoc?.defaultView ?? window;
 }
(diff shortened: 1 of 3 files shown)


packages/utilities/src/execution-context/getWindow.ts
```diff
@@ -14,5 +14,9 @@ export function getWindow(target: Event['target']): typeof window {
     return window;
   }
 
-  return target.ownerDocument?.defaultView ?? window;
+  let ownerDoc = target.ownerDocument;
```
istateside, line 17: This handles when `target` is the Document object itself

istateside:
@clauderic hey, do you have any interest in this patch for the legacy packages? I can close the PR if not, but this does handle some bugs that were impacting us in our usage of the library in an iframe

istateside:
Closing this so it's not hanging around in my open pull requests forever

PR #2020, opened 2026-04-17
fix(feedback): prevent DragOverlay flickering after drop

### LINKED ISSUE #1996: Flickering while using DragOverlay
Hi there, 

while using DragOverlay, there is some flickering when the dragged object is dropped down. This is also visible in dnd-kit examples: [https://main--5fc05e08a4a65d0021ae0bf2.chromatic.com/?path=/story/react-draggable-drag-overlay--drag-overlay](url)

Image

namgi2386:
Fixes #1996

## Problem

When using `DragOverlay`, a brief flicker occurs after dropping an item.
The overlay momentarily appears unstyled (at its natural/full size) before disappearing.

## Root Cause

The flicker is caused by a timing gap between style removal and React's re-render:

1. `cleanup()` calls `styles.reset()` — removes all CSS custom properties from the overlay element
2. `source.status = 'idle'` is set immediately after
3. This synchronously triggers `dragOperation.reset()` → `status = Idle`
4. `StyleInjector` reacts synchronously and removes the injected CSS rules from the document
   (including `[data-dnd-overlay]:not([data-dnd-dragging]) { display: none }`)
5. The browser paints at this point — the overlay has no inline styles and no CSS rules,
   but React hasn't re-rendered yet to remove the children
6. React re-renders on the next task and removes the children

The existing CSS rule `[data-dnd-overlay]:not([data-dnd-dragging]) { display: none }` was
intended to handle this, but it gets removed from the document in the same synchronous
execution before the browser has a chance to paint with it applied.

## Fix

Defer `source.status = 'idle'` (and subsequent finalization) to after the browser paint
by wrapping it in `setTimeout(fn, 0)` when operating in overlay mode.

This keeps the injected CSS rules — including
`[data-dnd-overlay]:not([data-dnd-dragging]) { display: none }` — alive in the document
at paint time, allowing the browser to apply them and hide the overlay before React
re-renders to remove the children.

`setTimeout` is the only deferral mechanism that executes after the browser paint;
microtasks and `requestAnimationFrame` both fire before paint.

Only applied when `feedbackElement === this.overlay` to avoid affecting non-overlay drag behavior.
No inline styles are used.

## Changes

`packages/dom/src/core/plugins/feedback/Feedback.ts`

- In `cleanup()`: extract finalization logic into a `finalize()` function; call it via
  `setTimeout(finalize, 0)` when in overlay mode, or call it directly otherwise

## Verification

1. Run storybook: `cd apps/stories && bun run dev`
2. Open: http://localhost:6006/?path=/story/react-draggable-drag-overlay--drag-overlay
3. Drag and drop the item — the overlay no longer flickers on drop

---

This is my first contribution to dnd-kit. Happy to make any changes if needed.

### DIFF
--- a/packages/dom/src/core/plugins/feedback/Feedback.ts
+++ b/packages/dom/src/core/plugins/feedback/Feedback.ts
@@ -89,9 +89,9 @@ export class Feedback extends Plugin<DragDropManager, FeedbackOptions> {
   constructor(manager: DragDropManager, options?: FeedbackOptions) {
     super(manager, options);
 
-    const styleInjector = manager.registry.plugins.get(
-      StyleInjector as any
-    ) as StyleInjector | undefined;
+    const styleInjector = manager.registry.plugins.get(StyleInjector as any) as
+      | StyleInjector
+      | undefined;
 
     const unregisterStyles = styleInjector?.register(CSS_RULES);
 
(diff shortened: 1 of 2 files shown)


clauderic:
>The existing CSS rule [data-dnd-overlay]:not([data-dnd-dragging]) { display: none } was
intended to handle this, but it gets removed from the document in the same synchronous
execution before the browser has a chance to paint with it applied.

Can you try solving it at this layer instead of applying inline styles? The library intentionally avoids applying any inline styles as much as possible.

namgi2386:
Thanks for the review! Agreed, inline styles aren't the right layer. I'll rework the fix at the stylesheet layer and push an update.

namgi2386:
Thank you for the feedback! 

I've updated the fix to avoid inline styles. Instead of setting `style.display = 'none'`, I now defer `source.status = 'idle'` (and the rest of the cleanup finalization) to after the browser paint using `setTimeout(fn, 0)`.

This keeps the injected CSS rule `[data-dnd-overlay]:not([data-dnd-dragging]) { display: none }` alive in the document at paint time, so the browser can apply it to hide the overlay before React re-renders to remove the children.

packages/dom/src/core/plugins/feedback/Feedback.ts
```diff
@@ -461,29 +461,37 @@ export class Feedback extends Plugin<DragDropManager, FeedbackOptions> {
...
+        ) {
+          placeholder.replaceWith(feedbackElement);
+        }
 
-      placeholder?.remove();
+        placeholder?.remove();
+      };
+
+      if (feedbackElement === this.overlay) {
+        setTimeout(finalize, 0);
```
clauderic, line 491: Would something like `await manager.renderer.rendering` have worked here?

namgi2386: Thank you! I'll look into it and update the PR

namgi2386: Thanks for the suggestion! I looked into whether `manager.renderer.rendering` 
could work here, but I believe it wouldn't reliably fix the flicker.

At the point where `cleanup()` is called (inside `animateTransform().then()`), 
there is no React rendering in progress — `rendering.current` is `null`, so 
`manager.renderer.rendering` returns `Promise.resolve()`. This means the 
`.then(finalize)` callback would run as a microtask, which executes *before* 
the browser paint — the same timing issue as calling `finalize()` directly.

The flicker occurs because the injected CSS rules are removed (synchronously, 
via `StyleInjector`) before the browser has a chance to paint with them applied. 
To fix this, `finalize()` — which triggers `source.status = 'idle'` and 
ultimately the CSS rule removal — needs to be deferred until *after* paint.

In the browser event loop: Task → Microtask → rAF → Paint → next Task (setTimeout)

`setTimeout(fn, 0)` defers execution to after paint (unlike microtasks or 
`requestAnimationFrame`, which both run before paint). I don't see an existing abstraction in the codebase for post-paint 
deferral, so I think `setTimeout` is the right tool here.

That said, happy to explore other approaches if you have something specific in mind!

PR #2026, opened 2026-04-19
fix(docs): correct wrong hook name references and descriptions in use-droppable

AlexRixten:
The useDroppable documentation contained several inaccuracies:

- element option: referenced useDraggable instead of useDroppable, and described the element as "draggable source" instead of "droppable target"
- collisionDetector option: missing "that" in "function can be used"
- effects option: referenced useDraggable instead of useDroppable
- isDropTarget output: described as "being dragged" instead of "a drop target"


clauderic:
Thanks!

PR #2027, opened 2026-04-23
fix(sortable): adding a guardrail to handle the null value that cause a TypeError in SortableContext

sourabh945:
## Problem
When the `items` array passed to `SortableContext` contains `null` values,
the component throws a TypeError because `typeof null === 'object'` in JS,
and `'id' in null` throws.

## Root Cause
typeof null returns "object" in JavaScript which causes the isObject()
check to pass, and then "id" in null throws a TypeError.

## Fix
Added a null check before the isObject() check to guard against this case.

## Reproduction
const items = [null, 'a', 'b'];
<SortableContext items={items}>...</SortableContext>
// TypeError: Cannot use 'in' operator to search for 'id' in null

### DIFF
--- a/.changeset/tender-aliens-build.md
+++ b/.changeset/tender-aliens-build.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/helpers': patch
+---
+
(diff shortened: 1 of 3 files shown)


clauderic (approved):
Thank you!

PR #2046, opened 2026-05-08
fix(sortable): Add support for non-contiguous indexes in OptimisticSortingPlugin

aidenfoxx:
## Problem
The OptimisticSortingPlugin is incompatible with non-contiguous sortable indices e.g. `items.map((id, i) => id !== "ignore" ?  <Sortable id={id} index={i} />  : null)`. When this happens, the plugin fails silently and prevents visual feedback for the dragged element.

This PR improves the ergonomics of the plugin to align with the non-optimistic behavior of dnd-kit.

Issue example: https://codesandbox.io/p/sandbox/tender-nash-pz8xfg

## Solution

The issue occurs when the code checks if the sortables have changed between microtask execution:
- The code uses the index of the generated `Set<Sortable>` to determine if  `sortable.index` has changed.
- This requires the sortables to have strict contiguous indexes starting at 0: `0,1,2,3`.

The PR solves the issue by taking a snapshot of the sortable indices before the microtask, and using those to determine if the index changed.

Additionally, the code in `dragend` was updated to resolve potentially non-contiguous `source.initialIndex` to an index relative to the generated contiguous set, so we can correctly determine the `target`.

**Before:**
Kooha-2026-05-08-10-22-10

**After:**
Kooha-2026-05-08-10-23-40

## Notes
- This is a non-breaking change.

### DIFF
--- a/packages/dom/src/sortable/plugins/OptimisticSortingPlugin.ts
+++ b/packages/dom/src/sortable/plugins/OptimisticSortingPlugin.ts
@@ -1,15 +1,17 @@
-import {Plugin, type UniqueIdentifier} from '@dnd-kit/abstract';
+import {Plugin} from '@dnd-kit/abstract';
 import type {DragDropManager} from '@dnd-kit/dom';
 import {move} from '@dnd-kit/helpers';
 import {batch} from '@dnd-kit/state';
 
 import {Sortable, SortableDroppable} from '../sortable.ts';
 import {isSortable} from '../utilities.ts';
+import {
+  getSortableIndices,
+  hasChanged,
+  type SortableInstances,
+} from './OptimisticSortingPlugin.helpers.ts';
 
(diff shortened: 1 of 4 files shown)


packages/dom/src/sortable/plugins/OptimisticSortingPlugin.ts
```diff
@@ -34,11 +34,41 @@ export class OptimisticSortingPlugin extends Plugin<DragDropManager> {
         }
       }
 
-      for (const [group, instances] of sortableInstances) {
-        sortableInstances.set(group, new Set(sort(instances)));
```
aidenfoxx, line 38: **note:** Sorting here is no longer required since the updated `hasChanged` logic has no dependency on the order of the sortable set.

clauderic (approved):
Thank you!

PR #2059, opened 2026-05-18
feat(react): re-export core utilities as namespaces (closes #2025)

### LINKED ISSUE #2025: Re-export utilities from framework packages
A lot of times, one wants to use utilities from the "core" packages, e.g. some collision detection customization from `@dnd-kit/collision` or some plugin customization from `@dnd-kit/dom` or `@dnd-kit/abstract` for types used in event handlers. It is good practice to directly depend on packages that are being imported directly (there are also lint rules for that). So now i need to add multiple `@dnd-kit/*` deps to the package.json:

```json
{
  "dependencies": {
    "@dnd-kit/react": "^0.3.0",
    "@dnd-kit/dom": "^0.3.0",
    "@dnd-kit/collision": "^0.3.0",
    "@dnd-kit/abstract": "^0.3.0"
  }
}
```

It would be nice if most of those could be re-exported from the framework entrypoint (in the above example react). One way would be to use ESM namespaces (which can be tree-shaken properly in modern build chains):
```ts
// @dnd-kit/react index.ts
export * as collision from '@dnd-kit/collision';

// MyComponent.tsx
import { useDroppable, collision } from '@dnd-kit/react';

useDroppable({ collisionDetector: collision.pointerDistance })
```

timagixe:
## Summary

Closes #2025.

Adds namespaced re-exports from `@dnd-kit/react`:

- `abstract` from `@dnd-kit/abstract`
- `collision` from `@dnd-kit/collision`
- `dom` from `@dnd-kit/dom`

This lets React consumers access common lower-level utilities without adding each package as a direct application dependency.

## Notes

This is intentionally React-only for the first pass, matching the issue label. The same pattern can be mirrored to other framework adapters if maintainers prefer.

This also keeps the re-exports namespaced rather than using unnamespaced `export *`, to avoid expanding the React root API with every lower-level export.

## Verification

- `bun run build` in `packages/react`
- Runtime smoke import for `abstract`, `collision`, and `dom`


clauderic (commented):
## Summary
Adds namespaced re-exports `abstract`, `collision`, and `dom` to `@dnd-kit/react`, so consumers can pull collision detectors, plugins, and types through the React entry point without each `@dnd-kit/*` package becoming a direct app dependency. Matches the request in #2025.

## Feedback

### Blocking issues
None.

### Suggestions
- `packages/react/src/core/index.ts:28` — namespace re-exports rather than `export *` is the right call. It avoids future name collisions when lower-level packages add new symbols, and consumers still get tree-shaking with modern bundlers. Good judgment in the PR notes on this.
- `packages/react/package.json:67` — `@dnd-kit/collision` is added as a direct dependency. That's appropriate since it's now imported at top level (previously it was reachable only transitively through `@dnd-kit/dom`).
- `packages/react/tsup.config.ts:7` — `@dnd-kit/collision` correctly added to `external` so it isn't bundled.
- One thing for the maintainer to weigh: this is a public API surface expansion (the `abstract`, `collision`, `dom` namespace names are now part of the contract). The PR is intentionally React-only for a first pass — if the same pattern lands in `@dnd-kit/vue` / `@dnd-kit/svelte` / `@dnd-kit/solid` later, naming consistency across adapters is worth keeping in mind.

### Changeset
- Status: present, `minor` for `@dnd-kit/react`. Correct — this is a new public API.

## Overall
Clean, minimal change. Thanks for following the user's exact example from #2025 and for the thoughtful note explaining the namespaced approach.

---
[claude-review]

PR #2061, opened 2026-05-19
fix(abstract): recompute transforms for drag operation snapshots (closes #2055)

### LINKED ISSUE #2055: `event.operation.transform` never gets updated when feedback is "none"
**Issue**:

When feedback is disabled via `Feedback.configure({ feedback: 'none' })`, `operation.transform` in the `onDragStart` / `onDragMove` / `onDragEnd` events always remains at its initial value `{ x: 0, y: 0 }`.

**Background**:

I'm using dnd-kit to implement card resize handles. In this scenario, I don't need any feedback element or overlay at all — the handle will be positioned according to the card's size rather than the mouse position.

In this case, I rely on the `transform` value from the events to compute the latest card's size.

**Root cause**:

`dragOperation.snapshot().transform` only reads from the cache without triggering a recomputation, and the `transform` is only computed inside the Feedback plugin (and only when a feedback element exists). 

As a result, when `feedback === 'none'`, `dragOperation.snapshot().transform` will always return only the initial value.

This looks like a bug — having system event values depend heavily on the internal behavior of a plugin feels like a leaky abstraction.

timagixe:
## Summary

Closes #2055.

This fixes drag operation event snapshots so `event.operation.transform` is kept up to date even when visual feedback is disabled with `Feedback.configure({feedback: 'none'})`.

The root cause was that `DragOperation.snapshot()` returned the cached transform, while the cache was normally refreshed as a side effect of the `Feedback` plugin reading `dragOperation.transform`. When feedback was disabled, that side effect never happened.

## Changes

- Recompute transforms in `DragOperation.snapshot()` instead of relying on a stale cached value.
- Snapshot proposed coordinates for `dragmove` events before the move is committed, while preserving `preventDefault()` behavior.
- Preserve same-tick relative move accumulation for queued `by` moves.
- Add abstract regression tests for transform snapshots, modifiers, prevented moves, and queued relative moves.
- Add DOM coverage for `Feedback.configure({feedback: 'none'})`.
- Add a patch changeset for `@dnd-kit/abstract`.

## Testing

- `bun test packages/abstract/tests`
- `bun test packages/dom/tests`
- `bunx tsc --noEmit -p packages/abstract/tsconfig.json`

## Notes

No stories or docs were added because this is a bug fix to existing event snapshot behavior, not a new feature or API.

### DIFF
--- a/.changeset/fast-wolves-reply.md
+++ b/.changeset/fast-wolves-reply.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/abstract': patch
+---
+
+Fix drag operation snapshots to report current transforms when feedback is disabled.
--- a/packages/abstract/src/core/manager/actions.ts
+++ b/packages/abstract/src/core/manager/actions.ts
@@ -189,10 +189,12 @@ export class DragActions<
         return;
       }
 
-      const coordinates = args.to ?? {
-        x: dragOperation.position.current.x + (args.by?.x ?? 0),
-        y: dragOperation.position.current.y + (args.by?.y ?? 0),
-      };
+      const getCoordinates = () =>
+        args.to ?? {
+          x: dragOperation.position.current.x + (args.by?.x ?? 0),
+          y: dragOperation.position.current.y + (args.by?.y ?? 0),
+        };
+      const coordinates = getCoordinates();
 
       const event = defaultPreventable(
         {
@@ -213,7 +215,7 @@ export class DragActions<
           return;
         }
 
-        dragOperation.position.current = coordinates;
+        dragOperation.position.current = getCoordinates();
       });
     });
   }
--- a/packages/abstract/src/core/manager/actions.ts
+++ b/packages/abstract/src/core/manager/actions.ts
@@ -189,10 +189,15 @@ export class DragActions<
         return;
       }
 
+      const coordinates = args.to ?? {
+        x: dragOperation.position.current.x + (args.by?.x ?? 0),
+        y: dragOperation.position.current.y + (args.by?.y ?? 0),
+      };
+
       const event = defaultPreventable(
         {
           nativeEvent: args.event,
-          operation: dragOperation.snapshot(),
+          operation: dragOperation.snapshot(coordinates),
           by: args.by,
           to: args.to,
         },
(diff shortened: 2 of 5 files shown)


clauderic (commented):
## Summary
Reworks `DragOperation.snapshot()` and `DragActions.move()` so the snapshot passed to `dragmove`/`dragend` listeners reflects the **proposed** coordinates and the transform derived from running them through the modifier pipeline, rather than the stale post-microtask values. Adds focused regression tests in [packages/abstract/tests/drag-operation-snapshot.test.ts](packages/abstract/tests/drag-operation-snapshot.test.ts) and [packages/dom/tests/feedback-none-transform.test.ts](packages/dom/tests/feedback-none-transform.test.ts), including a `preventDefault()` case verifying the rejected coordinates are not committed.

## Feedback

### Suggestions
- `packages/abstract/src/core/manager/operation.ts:178` — `snapshot()` now has a side effect: when the passed `coordinates` equal `position.current`, it writes back into `this.#transform`. This couples snapshot creation to internal cache maintenance, which is subtle. A reader of `snapshot()` would not expect mutation. Consider lifting the cache update into the `transform` getter (which already writes `#transform`) or into `move()` after the event isn't prevented. Not blocking — the behavior is correct, just non-obvious.
- `computeTransform()` now ignores `position.delta` in favor of `coordinates - position.initial`. That's equivalent today but worth a one-line comment noting the change of basis, since `position.delta` was the previous source of truth and someone might wonder why it's no longer used.
- The new modifier pipeline pass wraps each `createSnapshot(...)` in `untracked(...)`. The pre-existing code wrapped the *whole* iteration in `untracked` via the surrounding `snapshot()` body — confirming the modifier `apply()` cannot synchronously trigger a reactive read that needs to be tracked. Looks fine, but worth a sanity check.
- `packages/dom/tests/feedback-none-transform.test.ts:24` — casting `draggable` through `unknown` to mutate `element` is a smell; the test would be cleaner constructing the `Draggable` with an `element` option (if available) or using a real `HTMLDivElement` via JSDOM.

### Changeset
- Status: present and correct (`'@dnd-kit/abstract': patch`).

## Overall
Solid fix to a real foot-gun — the assertions for the `preventDefault` and accumulating `by` cases are especially nice. The snapshot side-effect is worth a maintainer eye but I don't think it blocks the fix. Thanks @timagixe!

---
[claude-review]

PR #2064, opened 2026-05-21
fix(dom): bubble drop target state to droppable ancestors (closes #1812)

### LINKED ISSUE #1812: [experimental] isDropTarget is not bubbling to Droppable when it's true for Sortables above
Assuming the below structure:
```tsx
<DragDropProvider>
   <Droppable>
      {list.map(item => 
         <Sortable id={item.id}/>
      )}
  </Droppable>
</DragDropProvider>
```
when dragging over a Sortable, its `isDropTarget` is `true` but the Droppable's `isDropTarget` is `false`. 

This is a problem you can also see in the official examples over https://next.dndkit.com/react/guides/multiple-sortable-lists, especially in https://next.dndkit.com/react/guides/multiple-sortable-lists#moving-items-between-columns (where background color style is introduced)

Notice how you have to place the element between the sorting list and the droppable element for the droppable to change color, since `isDropTarget` isn't bubbled.

Image

timagixe:
## Summary

Fixes `isDropTarget` so DOM droppable ancestors become active when a nested droppable is the current drop target.

Closes #1812 

## Details

- Bubbles drop target state through DOM ancestry instead of collision candidates
- Keeps disabled and non-accepting droppables inactive
- Adds regression coverage for nested targets and unrelated colliding siblings

## Verification

- `bun test packages/dom/tests`
- `bun test packages/abstract/tests`
- `bun run build`

### DIFF
--- a/.changeset/fix-nested-droppable-target-state.md
+++ b/.changeset/fix-nested-droppable-target-state.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/dom': patch
+---
+
+Mark DOM droppable ancestors as drop targets when a nested droppable is the active target.
--- a/packages/dom/src/core/entities/droppable/droppable.ts
+++ b/packages/dom/src/core/entities/droppable/droppable.ts
@@ -5,7 +5,7 @@ import type {
 } from '@dnd-kit/abstract';
 import {defaultCollisionDetection} from '@dnd-kit/collision';
 import type {CollisionDetector} from '@dnd-kit/collision';
-import {reactive, signal, untracked} from '@dnd-kit/state';
+import {derived, reactive, signal, untracked} from '@dnd-kit/state';
 import type {BoundingRectangle, Shape} from '@dnd-kit/geometry';
 import {DOMRectangle, PositionObserver} from '@dnd-kit/dom/utilities';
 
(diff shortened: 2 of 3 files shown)


clauderic (commented):
## Summary
Adds a DOM-specific override of `isDropTarget` on [Droppable](packages/dom/src/core/entities/droppable/droppable.ts) so that ancestor droppables containing the currently active target also report as drop targets, as long as they accept the source and aren't disabled. Bundles a focused bun:test that exercises matching/rejected/disabled ancestors and target clearing.

## Feedback

### Suggestions
- `packages/dom/src/core/entities/droppable/droppable.ts:121` — using `@derived` here is good; just worth confirming reactivity flows through `manager.dragOperation.target` so consumers re-render when the target changes. The test reads `isDropTarget` synchronously after mutating `targetIdentifier`, which proves the value is correct but not necessarily that subscribers re-run. Not blocking — the existing `super.isDropTarget` is `@derived` too and presumably works.
- The early-return ordering (`disabled` before `source`/`accepts`/`target`) is fine, but consider whether `disabled` should also short-circuit `super.isDropTarget` — if a droppable is disabled, should it report as a target even when it *is* the current target? Probably out of scope for this PR.
- Consider a test where the source is a descendant of the parent droppable (i.e. dragging within a nested droppable) — the current logic would correctly mark the parent as a target. Worth covering but not required.

### Changeset
- Status: present and correct (`'@dnd-kit/dom': patch`).

## Overall
Nice fix and the test is well structured — covers the positive path, the rejection path, and disabling. Thanks for closing this one out, @timagixe!

---
[claude-review]

PR #2063, opened 2026-05-19
fix(docs): fix malformed navigation links in plugins and sensors section

albertonoys:
Replace relative href paths with absolute paths in plugins.mdx and sensors.mdx Card components to resolve incorrect URL generation.

I tested it with mintlify and it works correctly.

PR #2071, opened 2026-05-28
Update sensor links in documentation

ammarmbe:
Updated the `href` attributes in the sensor documentation cards to point directly to `./pointer-sensor` and `./keyboard-sensor` instead of the previous nested paths, which would lead to a 404 page.

### DIFF
--- a/apps/docs/docs/extend/sensors.mdx
+++ b/apps/docs/docs/extend/sensors.mdx
@@ -17,14 +17,14 @@ The `@dnd-kit/dom` package provides a set of built-in sensors that can be used t
   <Card
     title="Pointer"
     icon="arrow-pointer"
-    href="./sensors/pointer-sensor"
+    href="./pointer-sensor"
   >
     Handles mouse, touch, and pen input with configurable activation constraints.
   </Card>
   <Card
     title="Keyboard"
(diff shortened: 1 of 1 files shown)


clauderic (commented):
## Summary
Fixes the relative `href` values for the Pointer and Keyboard sensor cards on the [Sensors page](apps/docs/docs/extend/sensors.mdx). The previous `./sensors/pointer-sensor` resolves under `/extend/sensors/` to `/extend/sensors/sensors/pointer-sensor`, hence the 404. Using `./pointer-sensor` resolves correctly to `/extend/sensors/pointer-sensor`.

## Feedback

### Suggestions
- There's an overlapping PR #2063 that fixes the same 404s in `sensors.mdx` and also `plugins.mdx`, but uses absolute paths (`/extend/sensors/pointer-sensor`) instead. Worth coordinating — whichever is merged first will conflict the other. The relative approach here is more portable, but `plugins.mdx` has the same bug and would still need a separate fix.

### Changeset
- Status: not needed (docs-only change). 

## Overall
Thanks for the fix! Small, targeted, and correct. Just flagging the overlap with #2063 for the maintainer to decide which direction to take.

---
[claude-review]

PR #2073, opened 2026-05-28
Support scroll adjusted transform

### LINKED ISSUE #2069: [Feature Request] Support scroll adjusted translate
Hello,

In the legacy dnd-kit events, there was a `delta` field representing the scroll adjusted translate, whose value equals `viewport-based delta + scroll delta`. However the new version is missing this feature — currently there is only a `transform` field representing the viewport-based delta. 

I'm requesting that this functionality be ported to the new version.

I'm working a PR to port this feature referencing the legacy implementation.

jindong-zhannng:
Close https://github.com/clauderic/dnd-kit/issues/2069

Port scroll-adjusted translate functionality from legacy dnd-kit:

- Listen to scroll event to measure the scroll delta with the same algorithm as legacy implementation.
- Add `dragOperation.adjustedTransform`.
- Dispatch virtual dragmove event when scrolling.


clauderic (commented):
## Summary
Adds a new `ScrollCompensator` core plugin to `@dnd-kit/dom` and a `scrollAdjustment` / `adjustedTransform` field on `DragOperation`, restoring the scroll-adjusted `delta` semantics from the legacy `@dnd-kit/core`. While dragging, the plugin listens for scroll events on the source's owner document, filters to the source's scrollable ancestors, accumulates the delta against the initial scroll offsets, and dispatches a `virtual: true` move so listeners see the updated transform. Closes #2069.

## Feedback

### Suggestions
- `packages/abstract/src/core/manager/operation.ts:182` — `get adjustedTransform()` is decorated `@derived` but has a write side-effect (`this.#adjustedTransform = result`). The whole purpose of `#adjustedTransform` is to keep `snapshot()` in sync, but it only updates when something reads the getter. The `disposeAdjustedTransformEffect` in [ScrollCompensator.ts](packages/dom/src/core/plugins/scrolling/ScrollCompensator.ts:91) exists solely to *force* a read on every reactive change. That's two indirections to achieve what could be one line in `snapshot()`:
  ```ts
  adjustedTransform: {
    x: this.#transform.x + this.scrollAdjustment.x,
    y: this.#transform.y + this.scrollAdjustment.y,
  },
  ```
  This would let you drop both the `#adjustedTransform` cache and the second effect in the plugin. Worth simplifying before this lands — the current shape will be confusing to future readers.
- `packages/dom/src/core/plugins/scrolling/ScrollCompensator.ts:60` — `manager.actions.move({event, virtual: true})` dispatches `dragmove` with the **scroll event** as the `nativeEvent`. Consumers that branch on `nativeEvent` type (e.g. distinguishing pointer/keyboard) will see an unexpected `ScrollEvent`. Consider documenting this in the JSDoc for the `virtual` option in [actions.ts](packages/abstract/src/core/manager/actions.ts), or pass `event: undefined` if the scroll event is not semantically meaningful here.
- `packages/dom/src/core/plugins/scrolling/ScrollCompensator.ts:43` — `initial` is captured the first time the effect runs (when `dragging` flips true). If the source's `scrollableAncestors` change during a drag (e.g. the DOM structure shifts), neither the ancestor set nor `initial` are recomputed. Probably fine in practice — flagging in case it's intentional.
- No tests cover `ScrollCompensator` or `adjustedTransform` behavior. The legacy `delta` semantics were subtle (scroll inside the drag would update `delta` even without pointer movement) — a bun:test that starts a drag, simulates a scroll event on an ancestor, and asserts the dispatched `dragmove` carries the expected `adjustedTransform` would catch regressions.

### Changeset
- Status: present and correct (`'@dnd-kit/abstract': minor`, `'@dnd-kit/dom': minor`). Minor is right for a new public field/plugin export, pre-1.0.

## Overall
Thanks @jindong-zhannng — this is a thoughtful port of a feature that legacy users definitely rely on, and the docs update on [drag-drop-manager.mdx](apps/docs/docs/concepts/drag-drop-manager.mdx) is a nice touch. The main thing I'd ask for is simplifying the `adjustedTransform` plumbing so the cache and forcing-effect go away, plus a small test.

---
[claude-review]

jindong-zhannng:
Simplified `dragOperation.adjustedTransform` and added unit tests.

jindong-zhannng:
Hi @clauderic , can we move forward this PR?

PR #2074, opened 2026-05-29
Fix grouped record sorting with numeric group IDs

### LINKED ISSUE #2065: dnd-kit/helpers move fails if columns indentify using numbers
I followed https://dndkit.com/react/guides/multiple-sortable-lists/ but using my own data where both my columns and items have number IDs.

Got issues with duplicated items when I tried to use the move helper in onDragOver.

Turns out even though I consistently passed in columns using numbers everywhere, the implementation in move usessomething like:

```
const entries = Object.entries(items);
for (const [id, children] of entries) {
    ...
}
```

And then tried to compare id to what you pass in. Object.entries always makes the keys strings, so then the comparison fails.
`'240' === 240`

Took a while to figure out what was going on. Maybe worth mentioning in the docs, or just make it work for number IDs.

timagixe:
## Summary

Fixes grouped record handling in `move()` when sortable group IDs are numbers. Since object record keys are strings at runtime, numeric group IDs could fail parent/group comparisons and cause incorrect optimistic reconciliation, including duplicated items.

This change normalizes grouped record keys before comparing or indexing parent groups.

Closes #2065.

## Changes

- Normalize grouped record keys in `packages/helpers/src/move.ts`
- Fix numeric group handling for group-container targets
- Fix optimistic same-group reconciliation when group IDs are numbers
- Add regression tests for numeric grouped records
- Add a patch changeset for `@dnd-kit/helpers`

## Testing

- `bun test packages/helpers/tests/move.test.ts`
- `bun run --cwd packages/helpers build`

### DIFF
--- a/.changeset/proud-socks-film.md
+++ b/.changeset/proud-socks-film.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/helpers': patch
+---
+
+Fix grouped record sorting with numeric group IDs
--- a/packages/helpers/src/move.ts
+++ b/packages/helpers/src/move.ts
@@ -48,6 +48,19 @@ export function arraySwap<T extends any[]>(
 type Items = UniqueIdentifier[] | {id: UniqueIdentifier}[];
 
 /**
+ * Resolve a grouped record key from a UniqueIdentifier.
+ * Object keys are strings at runtime, even when sortable groups use numeric IDs.
+ */
+function getRecordKey(
+  items: Record<UniqueIdentifier, Items>,
+  id: UniqueIdentifier
+) {
+  const key = String(id);
+
+  return Object.prototype.hasOwnProperty.call(items, key) ? key : undefined;
+}
+
+/**
  * Check if the source has sortable index properties via duck typing.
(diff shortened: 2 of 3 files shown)


clauderic (approved):
## Summary
Normalizes grouped-record keys in `move()` so that numeric sortable group IDs work correctly. `Object.entries(items)` always returns string keys, so direct identity comparisons (`source.group !== sourceParent`) and conditional branches that flow `target.id` back into `targetParent` could mix string/number values and produce incorrect optimistic reconciliations — including duplicated items.

## Feedback

### Blocking issues
None.

### Suggestions
- `packages/helpers/src/move.ts:53` — nice use of `Object.prototype.hasOwnProperty.call` for the safe lookup; this avoids accidentally matching inherited keys like `toString`.

### Changeset
- Status: present, correct bump (`patch` for `@dnd-kit/helpers`), correct package.

## Overall
Clean, targeted fix with good regression tests covering both the group-container targeting path and the same-group reconciliation path. The `getRecordKey` helper makes the intent explicit at every call site that touches record keys. Thanks for the careful write-up on the root cause in the PR description!

---
[claude-review]

clauderic (approved):

PR #2079, opened 2026-06-03
fix(dom): guard parseScale/parseTranslate against undefined transform…

### LINKED ISSUE #2078: parseScale/parseTranslate crash with "Cannot read properties of undefined (reading 'split')" on browsers without individual CSS transform properties (Chromium < 104)
### Package

  @dnd-kit/dom

  ### Describe the bug

  `parseScale()` and `parseTranslate()` call `.split(' ')` on the value of the
  individual CSS transform properties (`scale` / `translate`) read from
  `getComputedStyle(element)`, but only guard against the literal string
  `'none'`. On browsers that **do not support the individual transform
  properties** (`scale`, `translate`, `rotate` — Chromium < 104, released Aug
  2022), `getComputedStyle(element).scale` returns `undefined` rather than
  `'none'`, so the very first line crashes:

  ```
  TypeError: Cannot read properties of undefined (reading 'split')
      at parseScale (parseScale.ts:6)
      at parseTransform (parseTransform.ts)
      at new DOMRectangle (DOMRectangle.ts)
      ...
      at PositionObserver
  ```

  The function's type signature (`scale: string`) assumes the property always
  exists, which is not true on legacy engines.

  We see this ~1000+ times/day in production exclusively from old Chromium-based
  browsers (e.g. a vendor browser still on a pre-104 Chromium core). It fires
  from the throttled `PositionObserver` callback while dragging in a sortable
  list.

  ### Root cause

  `packages/dom/src/utilities/transform/parseScale.ts`

  ```ts
  export function parseScale(scale: string) {
    if (scale === 'none') {        // only guards 'none', not undefined
      return null;
    }
    const values = scale.split(' '); // 💥 scale is undefined on old engines
    ...
  }
  ```

  Same problem in `parseTranslate.ts` (`translate.split(' ')`).

  ### Suggested fix

  Treat a missing value the same as `'none'`:

  ```ts
  export function parseScale(scale?: string) {
    if (!scale || scale === 'none') {
      return null;
    }
    ...
  }

  export function parseTranslate(translate?: string) {
    if (!translate || translate === 'none') {
      return null;
    }
    ...
  }
  ```

  Happy to open a PR (have one ready).

  ### Reproduction

  Trigger a drag with `@dnd-kit/dom` in any engine where
  `CSS.supports('scale', '1') === false` (Chromium < 104). `getComputedStyle`
  returns `undefined` for `.scale` / `.translate`, and `parseScale` throws.

  ### dnd-kit version

  @dnd-kit/dom@0.4.0 (also present on `main` as of today)

silence717:
… values

`getComputedStyle(el).scale` / `.translate` return `undefined` on engines that do not support the individual `scale`/`translate` CSS transform properties (e.g. Chromium < 104). The existing guard only handled the string `'none'`, so `scale.split(' ')` threw "Cannot read properties of undefined (reading 'split')" during drag (via PositionObserver -> DOMRectangle -> parseTransform). Treat a missing value the same as `'none'`.

Closes #2078

### DIFF
--- a/packages/dom/src/utilities/transform/parseScale.ts
+++ b/packages/dom/src/utilities/transform/parseScale.ts
@@ -1,5 +1,5 @@
-export function parseScale(scale: string) {
-  if (scale === 'none') {
+export function parseScale(scale?: string) {
+  if (!scale || scale === 'none') {
     return null;
   }
 
--- a/.changeset/fix-parse-transform-undefined.md
+++ b/.changeset/fix-parse-transform-undefined.md
@@ -1,0 +1,5 @@
+---
+"@dnd-kit/dom": patch
+---
+
+Fix `TypeError: Cannot read properties of undefined (reading 'split')` in `parseScale`/`parseTranslate` on browsers that do not support the individual `scale`/`translate` CSS transform properties (Chromium < 104), where `getComputedStyle` returns `undefined` instead of `'none'`.
--- a/packages/dom/src/utilities/transform/parseTransform.ts
+++ b/packages/dom/src/utilities/transform/parseTransform.ts
@@ -10,9 +10,9 @@ export interface Transform extends Coordinates {
 }
 
 export function parseTransform(computedStyles: {
-  scale: string;
+  scale?: string;
   transform: string;
-  translate: string;
+  translate?: string;
 }): Transform | null {
   const {scale, transform, translate} = computedStyles;
   const parsedScale = parseScale(scale);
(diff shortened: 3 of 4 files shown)


packages/dom/src/utilities/transform/parseScale.ts
```diff
@@ -1,5 +1,5 @@
-export function parseScale(scale: string) {
-  if (scale === 'none') {
+export function parseScale(scale?: string) {
+  if (!scale || scale === 'none') {
```
timagixe, line 2: I wonder if it makes sense to update types here both for scale and translate, what do you think?

https://github.com/clauderic/dnd-kit/blob/d39a8ea7a48f75a3d8c585081b37a98879b34932/packages/dom/src/utilities/transform/parseTransform.ts#L12-L16

So scale and translate are both `propName?: string`

silence717: Thanks! Made scale and translate optional to match the parse functions. Updated!

clauderic (approved):

PR #2085, opened 2026-06-13
Add `shouldScroll` predicate to `AutoScroller` (closes #2070)

### LINKED ISSUE #2070: AutoScroller scrolls sibling draggables
Currently, AutoScroller will scroll any scrollable container at the pointer position, as long as one exists.

This behaves strangely in a sortable list where items are themselves scrollable: when you drag an item and hover over another sibling item, the inside of that sibling gets unexpectedly scrolled.

**Proposals**

Only scrolling the source's ancestor containers and droppables seems like a reasonable default behavior to me.

Alternatively, having the AutoScroller accept a `shouldScroll` predicate (signature `(scrollable, source) => boolean`) and letting consumers decide the trigger condition is also a good direction.

WDYT?

I'd be happy to contribute once we have a plan.

timagixe:
## PR Description

Closes #2070.

## Summary

- Adds `AutoScroller.configure({shouldScroll})` to let consumers filter scrollable auto-scroll candidates.
- Keeps existing default auto-scroll behavior unchanged.
- Documents the new option with a scrollable sortable-item example.
- Adds a controlled Storybook example for scrollable sibling items.
- Adds unit and e2e coverage for predicate behavior and the #2070 sibling-scroll case.
- Adds a changeset for `@dnd-kit/dom`.

## Verification

- run `test`
- run `test:e2e`
- run `build`

### DIFF
--- a/.changeset/beige-llamas-bake.md
+++ b/.changeset/beige-llamas-bake.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/dom': minor
+---
+
+Added a `shouldScroll` predicate option to `AutoScroller.configure()` for filtering auto-scroll candidates during drag operations.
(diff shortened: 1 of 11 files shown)


jindong-zhannng:
@timagixe looks good to me. @clauderic could you move it forward?

Issue #1747, opened 2025-06-23
[experimental] Node.removeChild: The node to be removed is not a child of this node


JannesMeyer:
I'm using `@dnd-kit/react/sortable` with multiple sortable lists (i.e. multiple groups) and the `OptimisticSortingPlugin`.

Whenever I move an item from one group to another and then try to commit the changes in `onDragEnd` by updating React state I get the following error:

```
DOMException: Node.removeChild: The node to be removed is not a child of this node
```

It seems that React doesn't like that the DOM was modified outside of React. Is there any way to undo the optimistic changes right before calling `setState()`?

The only two differences I see between my code and the [sandbox examples](https://codesandbox.io/p/sandbox/5qwvjf) are:

1. Different React version
2. I'm updating state in `onDragEnd` instead of `onDragOver`. Is this approach not supported?

My React version is `18.3.1` and @dnd-kit/react is `0.1.19`.

matheusAle:
I have the same issue and question ⏫

fantasy-zeng:
any solution?

matheusAle:
@fantasy-zeng,  this lib required us to optimistically add the drag item to the drop target droppable. Not doing so will cause  this error.

I needed to call a mutation only when the element drops, but onDragEnd will always have the source and the target as the same element, as you may have notice.
 
What I did was, create a ref with my update payload and change it on `DragDropProvider#onDragOver` callback and optimistically move the source to the target droppable when it is moved to a new one.

```ts
const dropUpdate = useRef<any>({});

<DragDropProvider
  onDragOver={({ operation }) => {
    const { target, source, shape, position } = operation;

    //? when we move the item to a new list, this callback is called again when source as the target 
    if (target.id === source.id) {
      return;
    }

   // ... check for source and target droppable groups
 }}

```

Boltawy:
@matheusAle Could you please clarify more ? I have the same issue.

Boltawy:
> Is there any way to undo the optimistic changes right before calling setState()?  


This.
My technical knowledge is limited, but I don't see why this isn't provided in a simple way by dnd-kit.
edit: formatting

clauderic:
Just call `event.preventDefault()` in `onDragOver` to prevent optimistic sorting from happening. Pretty sure this is documented, feel free to open a PR if you think the docs can be improved.

I’ll investigate the root issue when I can.

Boltawy:
[...]

Issue #1780, opened 2025-08-15
[experimental] useSortable breaks when dragging subitems from one group to another


valtism:
I've been trying to recreate a drag and drop solution with nested sortable items.

<img width="132" height="213" alt="Image" src="https://github.com/user-attachments/assets/630277e0-dcf3-4220-a1dd-6d994ae8b741" />

The data is in a nested structure, with items and subitems.

```json
[
  {
    "id": "1",
    "name": "Item 1",
    "data": [
      { "id": "3", "name": "Subitem 1" },
      { "id": "4", "name": "Subitem 2" }
    ]
  },
  {
    "id": "2",
    "name": "Item 2",
    "data": [
      { "id": "5", "name": "Subitem 3" },
      { "id": "6", "name": "Subitem 4" }
    ]
  }
]

```

Each time the list is re-sorted, I want to save the new order in `useState`.

## Reproduction

https://codesandbox.io/p/sandbox/7zl6wj

Sorting items works. Sorting subitems within an item works, but dragging a subitem to another item breaks react with the error message `Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.` in Webpack and `NotFoundError: The object can not be found here.` in Vite.

---

I can see that in the guide for [Multiple sortable lists](https://next.dndkit.com/react/guides/multiple-sortable-lists) reordering is done with a separate index array state. This could probably work, but I feel like my use-case is pretty common.

Furthermore, in production mode this all seems to work. I think that if there was a way to suppress this error things would be fine.

I've tried using `preventDefault` on onDragOver, but this loses a lot of the benefit to having the library handle sorting.

valtism:
FWIW I managed to get something working by handling all state changes in `onDragOver` and using `preventDefault` inside to manually handle the changes

clauderic:
Can you make sure to make the sandbox public?

<img width="659" height="506" alt="Image" src="https://github.com/user-attachments/assets/b98cb6a6-64c2-4f07-afd8-cc44136e2348" />

valtism:
@clauderic my apologies, it's public now

ChrisRM:
Having the exact same problem. Whenever I'm moving an item (subitem in your case) to a different column (item in your case), I get the same error.

I feel there should be an example in the documentation dealing with nested and/or separated arrays, as that is the most used structure for [multiple-sortable-lists](https://next.dndkit.com/react/guides/multiple-sortable-lists). 

[...]

PR #2102, opened 2026-07-04
fix(dom): restore DOM elements on dragend to prevent removeChild crash

ElitoSear:
### Summary
This PR fixes issues #1747 and #1780 where the application crashes with DOMException: Node.removeChild: The node to be removed is not a child of this node (or NotFoundError on Vite/Rollup) when dragging sortable items or elements across different groups/containers.

### Cause of the Issue
During dragging operations, OptimisticSortingPlugin and Feedback directly manipulate the DOM tree (e.g., using insertAdjacentElement to reorder nodes, or appending dragging elements to a popover/overlay root element). When a drag operation ends, the framework (React/Solid/Vue/Svelte) updates state and reconciles the DOM tree. However, because the DOM structure was altered directly outside the framework, the virtual DOM reconciliation gets out of sync, leading to errors like parent.removeChild failing to find the target element.

### Solution
1. **OptimisticSortingPlugin**:
   - Capture the original parent and sibling elements of the dragging element on dragstart.
   - On dragend, always restore the dragging element back to its original parent and position synchronously before the framework reconciles the DOM.
2. **Feedback**:
   - Detect if the feedback element is the real element itself (i.e. not using a separate DragOverlay).
   - If not using a separate DragOverlay, bypass the asynchronous drop animation and synchronously clean up / restore the element's original position on dragend to ensure a clean DOM state for the renderer's reconciliation tick.

This completely avoids any DOM reconciliation mismatches and resolves the crashes across all list component structures.

### DIFF
--- a/.changeset/fix-dom-reconciliation-crash.md
+++ b/.changeset/fix-dom-reconciliation-crash.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/dom': patch
+---
+
+Restore elements to their original DOM position on drag end to prevent framework DOM reconciliation crash (e.g. Node.removeChild error).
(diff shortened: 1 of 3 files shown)


DJ1TJOO:
Any update on this?

Issue #2107, opened 2026-07-14
autoScroll causes continuous flickering when dragging to the right or bottom edge of a scrollable container


hiwurun:
## Description

When `autoScroll` is enabled, continuously dragging an item toward the right or bottom edge of a scrollable container (keeping the pointer in the auto-scroll activation zone) causes noticeable **flickering / jitter**.

## Steps to reproduce

1. Open the official example: https://tanstack.com/table/v8/docs/framework/react/examples/column-dnd
2. Start dragging any column header
3. Move the pointer to the right (or bottom) edge of the scrollable area and hold it there to trigger continuous auto-scrolling
4. Observe the flickering during the scroll

Reproduction: https://tanstack.com/table/v8/docs/framework/react/examples/column-dnd  
(Reproducible on the latest versions)

## Expected behavior

Auto-scrolling should be smooth without any visual flickering or jitter while dragging.

## Actual behavior

- Continuous flickering occurs when auto-scrolling to the right or downward
- The flickering persists as long as auto-scroll is active
- Setting `autoScroll={false}` completely eliminates the issue

## Environment

- `@dnd-kit/core`: ^6.3.1
- `@dnd-kit/sortable`: ^10.0.0
- `@dnd-kit/utilities`: ^3.2.2
- `@tanstack/react-table`: ^8.21.3
- Browser:
- OS:

## Additional context

This issue is particularly noticeable in table column reordering scenarios where the container has horizontal overflow.

clauderic:
Thanks for the report, and for pointing at a concrete page that reproduces it.

One thing to flag: the versions you've listed (`@dnd-kit/core@^6.3.1`, `@dnd-kit/sortable@^10.0.0`, `@dnd-kit/utilities@^3.2.2`) are the **legacy** dnd-kit packages. The current library is a ground-up rewrite published under `@dnd-kit/dom`, `@dnd-kit/react`, `@dnd-kit/abstract` etc., with a different auto-scroll implementation (`Scroller` / `AutoScroller` in `@dnd-kit/dom`). So a fix here wouldn't be a shared code path.

Labelling as a bug + performance for the maintainer to triage against the legacy packages. If you're in a position to try the current packages, it would be genuinely useful to know whether the flicker reproduces there too — that would tell us whether this is a legacy-only regression or something inherent to how the auto-scroll zone interacts with a horizontally-overflowing container.

PR #2109, opened 2026-07-24
fix(dom): reset auto scroller state on cleanup

### LINKED ISSUE #2072: Dragging a item with auto-scroll, can in certain situation cause stuck scroll
If you drag a item within a context with autoScroll and drop it before its done scrolling for me it causes the scrolling to be stuck and page continue to scroll inifite, this can be stacked where u have multiple scroll actions at once.

This mainly seem to be a issue if u have two contexts but still a little unsure if its the one time this can happen.

wanxiankai:
## Summary
- reset the shared Scroller.autoScrolling flag when AutoScroller cleans up its interval
- add a regression test for AutoScroller cleanup
- add a patch changeset for @dnd-kit/dom

Closes #2072.

## Testing
- ./node_modules/.bin/prettier --check packages/dom/src/core/plugins/scrolling/AutoScroller.ts packages/dom/tests/auto-scroller.test.ts .changeset/quiet-scrolls-stop.md
- ./node_modules/.bin/tsc packages/dom/tests/auto-scroller.test.ts --noEmit --target es6 --module NodeNext --moduleResolution NodeNext --lib dom,dom.iterable,ES2015 --allowImportingTsExtensions --isolatedModules --strict --skipLibCheck --types bun-types

Note: I could not run bun test locally because bun is not installed in this environment.

### DIFF
--- a/.changeset/quiet-scrolls-stop.md
+++ b/.changeset/quiet-scrolls-stop.md
@@ -1,0 +1,5 @@
+---
+'@dnd-kit/dom': patch
+---
+
+Stop AutoScroller from leaving the shared scroller in an auto-scrolling state after its interval is cleaned up.
--- a/packages/dom/src/core/plugins/scrolling/AutoScroller.ts
+++ b/packages/dom/src/core/plugins/scrolling/AutoScroller.ts
@@ -68,6 +68,7 @@ export class AutoScroller extends Plugin<DragDropManager, AutoScrollerOptions> {
 
           return () => {
             clearInterval(interval);
+            scroller.autoScrolling = false;
           };
         } else {
           scroller.autoScrolling = false;
--- a/packages/dom/tests/auto-scroller.test.ts
+++ b/packages/dom/tests/auto-scroller.test.ts
@@ -1,0 +1,37 @@
+import {describe, expect, it, jest} from 'bun:test';
+import type {DragDropManager} from '@dnd-kit/abstract';
+
+import {AutoScroller} from '../src/core/plugins/scrolling/AutoScroller.ts';
+import {Scroller} from '../src/core/plugins/scrolling/Scroller.ts';
+
+function createManager(scroller: Pick<Scroller, 'autoScrolling' | 'scroll'>) {
+  return {
+    dragOperation: {
+      position: {current: {x: 0, y: 0}},
+      status: {dragging: true},
+    },
+    registry: {
+      plugins: {
+        get(plugin: unknown) {
+          return plugin === Scroller ? scroller : undefined;
+        },
+      },
+    },
+  } as DragDropManager<any, any>;
+}
+
+describe('AutoScroller', () => {
+  it('resets the scroller state when auto-scrolling is cleaned up', () => {
+    const scroller = {
+      autoScrolling: false,
+      scroll: jest.fn(() => true),
+    };
+    const autoScroller = new AutoScroller(createManager(scroller));
+
+    expect(scroller.autoScrolling).toBe(true);
+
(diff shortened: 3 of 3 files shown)


clauderic (commented):
## Summary

Sets `scroller.autoScrolling = false` in the cleanup of `AutoScroller`'s effect, so the shared `Scroller` isn't left believing it's auto-scrolling after the interval is cleared. Adds a `bun:test` unit test and a changeset.

Thanks for spotting this and for bringing a test and a changeset along — the "leaves stale state behind on teardown" instinct is the right one.

## Feedback

### Concerns worth resolving before merge

- `packages/dom/src/core/plugins/scrolling/AutoScroller.ts:71` — **this cleanup runs on every re-render of the effect, not just on teardown.** The effect consumes `manager.dragOperation.position`, so it re-runs on every position change during a drag. The sequence per pointer move becomes: cleanup → `autoScrolling = false` → effect body → `scroller.scroll(...)` → `autoScrolling = true`. That means `scroll()` now always observes `autoScrolling === false` during continuous auto-scroll, which disables two guards in `Scroller` that specifically depend on it (`packages/dom/src/core/plugins/scrolling/Scroller.ts`):

  ```ts
  if (this.autoScrolling && previousScrollableElements &&
      elements.size < previousScrollableElements?.size) {
    return previousScrollableElements;   // keeps the scrollable set from shrinking mid-scroll
  }
  ```

  ```ts
  if (this.autoScrolling && previousScrollBy) {
    const scrollIntentMismatch = ...      // suppresses scroll-intent oscillation
  }
  ```

  Both exist to stabilise continuous auto-scroll, so neutralising them risks regressing the flicker/stuck-scroll class of issues (e.g. #2072, #2107). Scoping the reset to genuine teardown — rather than to every effect re-run — would avoid this. One option is to reset from `destroy()` / on `dragend` instead of from inside the reactive effect's cleanup.

- Related: `autoScrolling` is `@reactive`, and `scroller.scroll()` is called directly in the effect body, so reads of `autoScrolling` inside it are tracked as dependencies of this effect. Today the body only ever writes `true`, so the write settles after one extra pass. With a cleanup that writes `false`, the sequence becomes `false → true → false → true …`, which looks like it could churn or loop. We weren't able to run the suite to confirm, so please treat this as a question rather than a finding — but it's worth verifying with a real drag before merging.

### Suggestions

- `packages/dom/tests/auto-scroller.test.ts` — the fake manager uses plain objects for `dragOperation.position` / `status` rather than signals, so the effect never re-runs the way it does in production. That's exactly the path where the concerns above live, so this test would pass either way. If a reactive fixture is feasible, asserting that `autoScrolling` stays `true` across successive position updates would be the more valuable assertion.
- `packages/dom/tests/auto-scroller.test.ts:31` — `expect(scroller.autoScrolling).toBe(true)` immediately after construction relies on the effect running synchronously. A brief comment noting that would help.
- The changeset description reads well; consider mentioning the user-visible symptom (what goes wrong when the flag is left set) so it's meaningful in the changelog.

### Changeset

- Status: **present and correct** — `'@dnd-kit/dom': patch`, right package, right bump type for a bug fix.

## Overall

The diagnosis is plausible and the hygiene around the PR (test + changeset) is appreciated. The hesitation is purely about placement: putting the reset in a reactive effect's cleanup makes it fire on every pointer move, which changes `Scroller` behaviour well beyond the teardown case. Moving it to a true teardown path would likely address the original bug without that exposure. Leaving this as a comment rather than a block so @clauderic can weigh in — he has the most context on the `autoScrolling` guards.

---
[claude-review]

Issue #2112, opened 2026-07-29
[Svelte] DragDropProvider briefly initializes default sensors before applying the sensors prop, causing draggables to bind stale sensors permanently


eythaann:
Bug summary

When using <DragDropProvider sensors={...}> from @dnd-kit/svelte without also passing a pre-built manager prop, the internal manager is constructed as new DragDropManager({}) (empty input) in DragDropProvider.svelte. Because the constructor input has no sensors, DragDropManager falls back to defaultPreset.sensors ([PointerSensor, KeyboardSensor]) — regardless of what the caller passed via the sensors prop.

The custom sensors prop is only applied afterwards, inside a $effect:

const manager = managerProp ?? new DragDropManager({}); // <-- always default sensors here
```
$effect(() => {
  manager.sensors = resolveCustomizable(sensors, defaultPreset.sensors); // applied too late
});
```
This would be harmless if it only affected the manager's own sensor registry, but each Draggable/Sortable binds sensors at mount time, reading manager.sensors synchronously in its own registration effect (dom/index.js, Draggable effects block):

```
const sensors = this.sensors?.map(descriptor) ?? [...manager2.sensors];
const unbindFunctions = sensors.map((entry) => {
  const sensorInstance = ...;
  return sensorInstance.bind(this, options); // binds keydown listener for KeyboardSensor
});
```

If any draggable registers before the provider's $effect corrects manager.sensors, that draggable's KeyboardSensor binding (e.g. a real keydown listener on the handle) is created and persists — it is only removed when that draggable itself unmounts, not when manager.sensors is later updated. So even though the manager's sensor list is correctly fixed a moment later, the affected draggables keep a live keyboard-activation listener for their entire lifetime.

Observed impact

Passing sensors={[PointerSensor.configure(...)]} (deliberately excluding KeyboardSensor) does not actually disable keyboard-triggered drag activation — pressing Enter on a focused draggable handle still starts a drag operation, because the KeyboardSensor was bound before the corrected sensor list took effect.

Reproduction

1. Render <DragDropProvider sensors={[PointerSensor]} plugins={...}> with a Sortable/Draggable item inside, without passing a manager prop.
2. Focus a draggable handle and press Enter.
3. Observe that a drag operation starts, even though KeyboardSensor was never included in sensors.

Expected behavior

[...]

PR #2113, opened 2026-07-29
[#2112] Fix not cleanup of default sensors for Svelte library

eythaann:
Previously the manager was constructed with an empty config, so defaultPreset.sensors (including KeyboardSensor) was always instantiated and bound to draggables before the reactive effect corrected it. Destroying a sensor doesn't unbind listeners already bound to draggables, so the default KeyboardSensor kept handling events even when a custom or no keyboard sensor was passed. Now the manager is created with the actual props, matching the React and Vue providers.

### DIFF
--- a/packages/svelte/src/core/context/DragDropProvider.svelte
+++ b/packages/svelte/src/core/context/DragDropProvider.svelte
@@ -43,8 +43,8 @@
 
   const {renderer, trackRendering} = createRenderer();
 
-  // Create manager once; plugins/sensors/modifiers are synced reactively via $effect below
-  const manager = managerProp ?? new DragDropManager({});
+  const manager =
+    managerProp ?? new DragDropManager({plugins, sensors, modifiers});
 
   manager.renderer = renderer;
 


clauderic (changes requested):
## Summary

Fixes #2112 by constructing the Svelte `DragDropProvider`'s `DragDropManager` with the initial `plugins` / `sensors` / `modifiers` props instead of an empty object, closing the window where draggables could bind the default preset's sensors before the `$effect` corrected them.

We verified this against the source and the fix is correct:

- `DragDropManager`'s constructor already resolves customizables itself (`packages/dom/src/core/manager/manager.ts:42-46` calls `resolveCustomizable` on `input.plugins`/`sensors`/`modifiers`), so passing the raw props through is safe and `undefined` still falls back to `defaultPreset`.
- This brings the Svelte adapter in line with React, which has always constructed with `new DragDropManager(input)` (`packages/react/src/core/context/DragDropProvider.tsx:81`) and therefore never had this bug.
- The `$effect` on lines 53-57 remains correct for subsequent prop updates.

Nice, minimal fix — thank you for tracking down the root cause and writing it up so precisely in the issue.

## Feedback

### Blocking issues

- **Missing changeset.** This modifies `packages/svelte/src/core/context/DragDropProvider.svelte`, so it needs a `.changeset/*.md`:

  ```
  ---
  '@dnd-kit/svelte': patch
  ---

  Fixed `DragDropProvider` briefly initializing default sensors before applying the `sensors` prop, which caused draggables to permanently bind stale sensors.
  ```

### Suggestions

- The removed comment (`// Create manager once; plugins/sensors/modifiers are synced reactively via $effect below`) was useful context for why the `$effect` below exists. Consider keeping a trimmed version, e.g. `// Seed the manager with the initial props so children never bind default sensors; later updates are synced via the $effect below.`
- A regression test would be valuable if it's feasible in the Svelte test setup — something that renders `<DragDropProvider sensors={[PointerSensor]}>` with a draggable and asserts that a keyboard activation does *not* start a drag. Not a blocker if there's no existing harness for it.

### Changeset

- Status: **missing**. Should be `'@dnd-kit/svelte': patch`.

## Overall

The change itself is exactly right and we'd be happy to see it land — just needs the changeset. Thanks for both the detailed issue and the fix.

---
[claude-review]

PR #2115, opened 2026-07-31
fix(ci): align workflow Node.js version with package engines

MGPOCKY:
## Why
Package metadata declares `engines.node` / `volta.node` / `.nvmrc` as `22`, but CI workflows still pin `node-version` to a mismatched value.

That can make CI run on a runtime the project no longer supports (or skip the version the package actually targets).

## What changed
Update the affected workflow `node-version` pins so they satisfy `22` (using `22` where a concrete pin is needed).

- `.github/workflows/continous-release.yml`
  - `node-version: 20` → `node-version: 22`
- `.github/workflows/release.yml`
  - `node-version: '20'` → `node-version: '22'`

## Test plan
- [ ] Package `engines.node` / `volta.node` / `.nvmrc` is still `22`
- [ ] Updated workflows now use versions consistent with that constraint
- [ ] Relevant CI jobs on this branch look healthy

Found while auditing CI/package version consistency across popular repos.

### DIFF
--- a/.github/workflows/continous-release.yml
+++ b/.github/workflows/continous-release.yml
@@ -20,7 +20,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 22
           # https://github.com/atlassian/changesets/issues/550#issuecomment-811245508
           registry-url: https://registry.npmjs.org
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -30,7 +30,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
         with:
-          node-version: '20'
+          node-version: '22'
           # https://github.com/atlassian/changesets/issues/550#issuecomment-811245508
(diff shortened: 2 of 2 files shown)


clauderic (approved):
## Summary

Bumps `actions/setup-node` from Node 20 to Node 22 in both release workflows (`.github/workflows/release.yml` and `.github/workflows/continous-release.yml`).

## Feedback

We checked this against the repo: `.nvmrc` pins `22`, and those two workflows were the only places still requesting `20` (`tests.yml` and `playwright.yml` both use `oven-sh/setup-bun`, so they're unaffected). Both occurrences are updated, so the change is complete and consistent.

### Suggestions

- Minor: the title and description reference aligning with "package engines", but there's no `engines` field in `package.json` — the source of truth here is `.nvmrc`. Worth rewording so future readers don't go looking for a field that doesn't exist.

### Changeset

- Status: **not required**. Only `.github/workflows/` files are touched; nothing under `packages/*/src/` changed.

## Overall

Small, correct, and low-risk. Thanks for the cleanup — keeping CI in step with `.nvmrc` avoids a class of confusing release-time surprises.

---
[claude-review]

PR #2117, opened 2026-07-31
fix(react): defer provider manager cleanup

### LINKED ISSUE #2116: React 19: DragDropProvider manager is destroyed during Strict Mode replay
## Description

In React 19 development with `StrictMode` enabled, `DragDropProvider` can retain a manager instance after that instance has been destroyed during React’s development-only Effect replay.

`useStableInstance` creates the manager once and stores it in a ref, but its cleanup currently destroys `ref.current` from `useInsertionEffect`. The replayed lifecycle can therefore leave the mounted provider holding a destroyed manager.

## Reproduction

1. Use React 19 in development mode with `StrictMode` enabled.
2. Render a `DragDropProvider` containing a draggable element.
3. Interact with the draggable element after the initial mount.

```tsx
<StrictMode>
  <DragDropProvider>
    <Draggable />
  </DragDropProvider>
</StrictMode>
```

## Expected behavior

The manager should remain usable for as long as its `DragDropProvider` remains mounted.

## Actual behavior

The provider can retain a manager that has already been destroyed, so later drag interactions may fail.

## Proposed direction

Move manager disposal out of `useInsertionEffect` and defer cleanup enough to distinguish the development-only Strict Mode replay from a real unmount.

I have a focused patch with a Storybook regression story and Playwright coverage ready to link in a pull request.

mnlttt:
Fixes #2116

## Summary

`DragDropProvider` can destroy its manager during React Strict Mode’s development-only Effect replay.

This moves manager disposal out of `useInsertionEffect` and defers cleanup to a microtask. The version guard ignores the temporary Strict Mode cleanup while still destroying the manager after a real unmount.

- add an interactive Storybook regression story for provider cleanup
- add a Playwright regression test covering a keyboard drag followed by provider unmount

## Validation

- `bun run build`
- `playwright test apps/stories/tests/drag-drop-provider.spec.ts`

No changeset is included because this fixes lifecycle behavior without changing the public API.

### DIFF
--- a/apps/stories/stories/react/DragDropProvider/DragDropProvider.stories.tsx
+++ b/apps/stories/stories/react/DragDropProvider/DragDropProvider.stories.tsx
@@ -1,0 +1,43 @@
+import {useState} from 'react';
+import type {Meta, StoryObj} from '@storybook/react-vite';
+import {DragDropProvider, useDraggable} from '@dnd-kit/react';
+
+function Draggable() {
+  const {isDragging, ref} = useDraggable({id: 'draggable'});
+
+  return (
+    <button ref={ref} data-dnd-dragging={isDragging || undefined} type="button">
+      Draggable
+    </button>
+  );
+}
+
+function DragDropProviderCleanup() {
+  const [isMounted, setIsMounted] = useState(true);
+
+  return (
+    <div>
+      <button onClick={() => setIsMounted((mounted) => !mounted)} type="button">
+        {isMounted ? 'Unmount provider' : 'Mount provider'}
+      </button>
+      {isMounted ? (
+        <DragDropProvider>
+          <Draggable />
+        </DragDropProvider>
+      ) : (
+        <p>Provider unmounted</p>
+      )}
+    </div>
+  );
+}
+
+const meta: Meta<typeof DragDropProviderCleanup> = {
+  title: 'React/DragDropProvider',
+  component: DragDropProviderCleanup,
+  tags: ['!autodocs', 'hidden'],
+};
+
+export default meta;
+type Story = StoryObj<typeof DragDropProviderCleanup>;
+
+export const Cleanup: Story = {};
--- a/apps/stories/tests/drag-drop-provider.spec.ts
+++ b/apps/stories/tests/drag-drop-provider.spec.ts
@@ -1,0 +1,30 @@
+import {test, expect} from '../../stories-shared/tests/fixtures.ts';
+
+test.describe('DragDropProvider', () => {
+  test('cleans up without React errors', async ({dnd}) => {
+    const errors: string[] = [];
+
+    dnd.page.on('console', (message) => {
+      if (message.type() === 'error') {
+        errors.push(message.text());
+      }
+    });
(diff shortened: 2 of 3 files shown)


clauderic (changes requested):
## Summary

Fixes #2116 by moving `useStableInstance`'s disposal out of `useInsertionEffect` into a `useEffect` cleanup that defers `destroy()` to a microtask, guarded by a version counter so React 19 Strict Mode's development-only Effect replay doesn't destroy a manager that's still mounted.

We read `packages/react/src/core/context/DragDropProvider.tsx` and confirmed the bug you described: `ref.current` is only ever populated during render, so the insertion-effect cleanup destroyed the instance and the replayed setup had nothing to recreate it with. The version-counter approach is sound — on replay the setup bumps `cleanupVersion` before the queued microtask runs, so the stale cleanup no-ops; on a real unmount nothing bumps it and `destroy()` proceeds. Thanks for the clear write-up and for bringing coverage along with the fix.

## Feedback

### Blocking issues

- **Missing changeset.** `packages/react/src/core/context/DragDropProvider.tsx` is modified, so this needs a `.changeset/*.md`:

  ```
  ---
  '@dnd-kit/react': patch
  ---

  Fixed `DragDropProvider` retaining a destroyed manager during React Strict Mode's development-only Effect replay.
  ```

### Suggestions

- `apps/stories/tests/drag-drop-provider.spec.ts` — **this test doesn't appear to exercise the reported regression.** We couldn't find any `StrictMode` usage under `apps/stories/` (including `.storybook/preview.tsx`), so the story renders without the double-invoke that triggers the bug. The test also asserts on unmount behaviour ("cleans up without React errors"), whereas #2116 is about a *still-mounted* provider holding a destroyed manager. A test closer to the regression would wrap the story subtree in `<StrictMode>` and then assert that a drag still works *after* mount — e.g. pick up the draggable with the keyboard and assert `data-dnd-dragging` is set. As written, we'd expect this test to pass on `main` too, which would make it a non-regression test.

- `apps/stories/tests/drag-drop-provider.spec.ts:26` — `await dnd.page.waitForTimeout(0)` to flush the microtask is a bit implicit; a comment explaining what's being waited on would help future readers.

- `packages/react/src/core/context/DragDropProvider.tsx:153` — worth a short JSDoc `@remarks` on `useStableInstance` explaining *why* disposal is deferred to a microtask. The inline comment covers the "what" but the ordering assumption (Strict Mode re-runs setup synchronously, before microtasks flush) is subtle enough to be worth stating explicitly so it isn't "simplified" away later.

- Moving from `useInsertionEffect` to `useEffect` also moves disposal later in the commit phase for *real* unmounts. That's almost certainly fine here, but it is a behaviour change beyond the Strict Mode fix and worth a sentence in the PR description.

### Changeset

- Status: **missing**. Should be `'@dnd-kit/react': patch`.

## Overall

The core fix looks right and is well reasoned. Two things to land it: add the changeset, and either make the Playwright test actually reproduce the Strict Mode regression or reframe what it claims to cover. Thanks for the thorough report and for following it up with a patch.

---
[claude-review]

PR #2125, opened 2026-08-11
fix: remove popover attribute when showPopover fails to prevent Chrome hiding dragged item

### LINKED ISSUE #2111: Active sortable item becomes invisible during drag in Chrome
Description

While using @dnd-kit/svelte sortable, the active dragged item becomes invisible as soon as dragging starts.

Only the currently dragged item disappears. All other sortable items remain visible and continue updating correctly.

The drag operation itself works as expected:

Drag starts successfully
Collision detection works
Reordering works
The item is dropped in the correct position

The issue is only visual: the active dragged item is not rendered while dragging.

Environment
@dnd-kit/svelte: 0.5.0
@dnd-kit/helpers: 0.5.0
Svelte 5
Vite
Chrome 150.0.7871.187 (Stable)
Windows
Minimal reproduction

The issue reproduces with the minimal example from the documentation.

App.svelte

```
<script lang="ts">
    import { DragDropProvider } from '@dnd-kit/svelte';
    import { move } from '@dnd-kit/helpers';
    import SortableItem from './SortableItem.svelte';

    let items = $state(Array.from({ length: 20 }, (_, i) => i + 1));

    let snapshot: number[] = [];

    function onDragStart() {
        snapshot = items.slice();
    }

    function onDragOver(event: any) {
        items = move(items, event);
    }

    function onDragEnd(event: any) {
        if (event.canceled) {
            items = snapshot;
        }
    }
</script>

<DragDropProvider {onDragStart} {onDragOver} {onDragEnd}>
    <ul class="list">
        {#each items as id, index (id)}
            <SortableItem {id} {index} />
        {/each}
    </ul>
</DragDropProvider>

<style>
    .list {
        width: 300px;
        margin: 50px auto;
        padding: 0;
        list-style: none;
    }
</style>
```

SortableItem.svelte

```
<script lang="ts">
    import { createSortable } from '@dnd-kit/svelte/sortable';

    let { id, index } = $props();

    const sortable = createSortable({
        get id() {
            return id;
        },
        get index() {
            return index;
        },
    });
</script>

<li {@attach sortable.attach}>
    Item {id}
</li>
```

Actual behavior

The active draggable item becomes invisible immediately after dragging starts.

The item remains interactive and is dropped into the correct position, but it is not visible while dragging.

In DevTools the dragged element has:

```
<li
    data-dnd-dragging="true"
    popover="manual"
>
```

The browser applies the following user-agent rule:

```
[popover]:where(:not(:popover-open)) {
    display: none;
}
```

During dragging:

returns:

VIDEO
https://youtu.be/tGLkiSCbdXU

waterWang:
## Problem

In Chrome, when dragging a sortable item using `@dnd-kit/svelte` (or `@dnd-kit/dom`), the active dragged item becomes invisible. All other sortable items remain visible, and the drag operation itself works correctly (collision detection, reordering, drop).

The element has `popover="manual"` set but isn't `:popover-open`, so Chrome's UA stylesheet hides it:

```css
[popover]:where(:not(:popover-open)) { display: none; }
```

## Root cause

In `Feedback.ts`, the `popover="manual"` attribute is set unconditionally before calling `showPopover()`. However, `showPopover()` has guards that can fail silently:

- If the element is not connected to the DOM at that moment
- If `showPopover()` throws

When `showPopover()` fails, the `popover="manual"` attribute remains applied but the popover is never opened, triggering the Chrome UA rule that hides the element.

## Fix

Two changes:

1. **`showPopover.ts`**: Return a `boolean` indicating whether the popover was successfully shown (`true`) or not (`false`).

2. **`Feedback.ts`**: Check the return value of `showPopover()`. If it returns `false`, remove the `popover` attribute as a fallback, so the element stays visible.

## Test plan

1. Reproduce the issue with the minimal Svelte 5 sortable example from the issue report
2. Verify the dragged item is visible during drag in Chrome
3. Verify popover promotion still works when `showPopover()` succeeds (e.g. in cross-frame scenarios)
4. Verify no regression in the `Debug` plugin which also uses `showPopover()`

Fixes #2111

### DIFF
--- a/packages/dom/src/core/plugins/feedback/Feedback.ts
+++ b/packages/dom/src/core/plugins/feedback/Feedback.ts
@@ -363,7 +363,14 @@ export class Feedback extends Plugin<DragDropManager, FeedbackOptions> {
       if (!feedbackElement.hasAttribute('popover')) {
         feedbackElement.setAttribute('popover', 'manual');
       }
-      showPopover(feedbackElement);
+      if (!showPopover(feedbackElement)) {
+        // showPopover can fail silently if the element is not connected
+        // or showPopover() throws. When it fails, the popover="manual"
+        // attribute remains set but the popover is never opened, causing
+        // Chrome's UA stylesheet [popover]:not(:popover-open) { display: none }
+        // to hide the dragged element. Remove the attribute as a fallback.
+        feedbackElement.removeAttribute('popover');
+      }
       feedbackElement.addEventListener('beforetoggle', preventPopoverClose);
     }
 
--- a/packages/dom/src/utilities/popover/showPopover.ts
+++ b/packages/dom/src/utilities/popover/showPopover.ts
@@ -1,6 +1,6 @@
 import {supportsPopover} from './supportsPopover.ts';
 
-export function showPopover(element: Element) {
+export function showPopover(element: Element): boolean {
   try {
     if (
       supportsPopover(element) &&
@@ -10,8 +10,11 @@ export function showPopover(element: Element) {
       !element.matches(':popover-open')
     ) {
       element.showPopover();
+      return true;
     }
   } catch (error) {
     // no-op
   }
+
+  return false;
 }


MILLERMARRU (changes requested):
The diagnosis for the Chrome hiding bug is right, and I confirmed the UA rule (`[popover]:where(:not(:popover-open)) { display: none; }`) is exactly what fires when `popover="manual"` is set but the popover never actually opened.

But I think the fix as written introduces a real regression. Current `showPopover()` on `main`:

```ts
export function showPopover(element: Element) {
  try {
    if (
      supportsPopover(element) &&
      element.isConnected &&
      element.hasAttribute('popover') &&
      !element.matches(':popover-open')
    ) {
      element.showPopover();
    }
  } catch (error) {
    // no-op
  }
}
```

Note the last guard: `!element.matches(':popover-open')`. If the element is *already* open (e.g. `showPopover()` gets called a second time for the same feedback element, which can happen across re-renders or cross-frame promotion per the PR's own test plan item 3), this function intentionally skips calling `element.showPopover()` again and does nothing, since there's nothing to do.

With this PR's change, that "nothing to do because it's already open" path now returns `false` (it falls through to the final `return false` just like every other early-exit), which is indistinguishable from an actual failure. `Feedback.ts`'s new fallback (`if (!showPopover(feedbackElement)) { feedbackElement.removeAttribute('popover'); }`) can't tell the difference, so it strips the `popover` attribute off an element that's genuinely open and visible right now, which reintroduces the hiding bug through a different door (a mid-drag `showPopover()` call while already open would deopen a currently-fine feedback element).

Suggest checking `element.matches(':popover-open')` directly after the attempt in `Feedback.ts` instead of trusting `showPopover()`'s return value, or having `showPopover()` also return `true` when the element is already open (only `false` on the actual `try`/`catch` failure or the `isConnected`/`supportsPopover` guards). Happy to be wrong if there's a reason "already open" and "failed" are meant to collapse to the same fallback here, but from reading the guard as-is I don't see how the fallback distinguishes them.

PR #2131, opened 2026-09-01
fix: respect `transition: null` in framework bindings

### LINKED ISSUE #2077: Sortable framework bindings ignore `transition: null`
## Summary

Passing `transition: null` to the React, Vue, Solid, or Svelte sortable APIs should disable sortable reorder animations. Instead, each binding merges `defaultSortableTransition` before constructing or updating the core `Sortable` instance, so `null` is replaced by the default transition object and items still animate.

This appears to be a bug. The docs, public types, and core implementation all treat `null` as the sentinel for disabling sortable transitions.

### Documentation supports `null`

- `apps/docs/docs/react/hooks/use-sortable.mdx` documents the React `transition` option as nullable.
- `apps/docs/docs/concepts/sortable.mdx` documents the core Sortable input as `SortableTransition | null`.
- `apps/docs/docs/legacy/presets/sortable/use-sortable.mdx` says: “To disable transitions entirely, set the `transition` argument to `null`.”

The Solid, Vue, and Svelte hook/composable docs currently omit `| null` in their local parameter tables. However, their public package types inherit the core `SortableInput`, where `transition` is nullable. That looks like a docs gap, not evidence that `null` is unsupported.

### Public and core types support `null`

In `packages/dom/src/sortable/sortable.ts`:

```ts
export interface SortableInput<T extends Data> {
  /**
   * The transition configuration to use when the index of the sortable item changes.
   */
  transition?: SortableTransition | null;
}
```

The core class also stores a nullable transition:

```ts
transition: SortableTransition | null;
```

### Core behavior treats `null` as “do not animate”

The `Sortable` constructor only applies the default when `transition` is omitted:

```ts
constructor(
  {
    transition = defaultSortableTransition,
    ...input
  }: SortableInput<T>,
  manager: DragDropManager<any, any> | undefined
) {
  // ...
  this.transition = transition;
}
```

An explicit `null` should therefore remain `null`.

`Sortable.animate()` then exits early when `transition` is falsy:

```ts
if (!shape || !transition || (idle && !transition.idle)) {
  return;
}
```

So the core implementation already supports disabling sortable transitions with `transition: null`.

### React

`packages/react/src/sortable/useSortable.ts` eagerly merges defaults:

```ts
const transition = {...defaultSortableTransition, ...input.transition};
```

When `input.transition` is `null`, the default object remains intact, so the core `Sortable` instance receives the default transition instead of `null`.

### Vue

`packages/vue/src/sortable/useSortable.ts` has the same issue:

```ts
const transition = computed(() => ({
  ...defaultSortableTransition,
  ...toValue(input.transition),
}));
```

### Solid

`packages/solid/src/sortable/useSortable.ts` merges defaults during construction:

```ts
const transition = {
  ...defaultSortableTransition,
  ...input.transition,
};
```

It also resets falsy transition values back to defaults during sync:

```ts
sortable.transition = input.transition
  ? {...defaultSortableTransition, ...input.transition}
  : defaultSortableTransition;
```

This prevents `null` from being preserved.

### Svelte

`packages/svelte/src/sortable/createSortable.svelte.ts` merges defaults during construction and sync:

```ts
transition: {
  ...defaultSortableTransition,
  ...input.transition,
},
```

```ts
sortable.transition = {
  ...defaultSortableTransition,
  ...input.transition,
};
```

## Reproduction

Stackblitz: https://stackblitz.com/edit/vitejs-vite-ypeenai9

Using React as a minimal example:

```tsx
import {useSortable} from '@dnd-kit/react/sortable';

function Item({id, index}: {id: string; index: number}) {
  const {ref, sortable} = useSortable({
    id,
    index,
    transition: null,
  });

  console.log(sortable.transition);

  return <div ref={ref}>{id}</div>;
}
```

Expected console output:

```text
null
```

Actual console output:

```json
{
  "duration": 250,
  "easing": "cubic-bezier(0.25, 1, 0.5, 1)",
  "idle": false
}
```

This can also be reproduced visually by passing `transition: null` to each `useSortable` call in a sortable list story, then dragging an item so another sortable item is displaced. The displaced items still animate, even though `null` should disable the sortable transition.

## Expected behavior

When users pass `transition: null`, the framework binding should pass `null` to the core `Sortable` instance. `Sortable.animate()` should then skip transition animation.

## Actual behavior

The framework binding converts `transition: null` into the default transition object, so sortable items still animate with the default settings.

## User impact

Users cannot disable sortable transitions through the documented and typed public API. This makes it harder to build:

- instant-reorder lists;
- reduced-motion variants;
- custom animation systems;
- deterministic no-animation tests and demos.

## Suggested fix

Preserve `null` before merging defaults:

```ts
function resolveSortableTransition(
  transition: SortableTransition | null | undefined
): SortableTransition | null {
  return transition === null
    ? null
    : {...defaultSortableTransition, ...transition};
}
```

Use the resolved value in the React, Vue, Solid, and Svelte bindings.

This preserves the intended behavior for all cases:

- `transition: undefined` or omitted uses defaults.
- `transition: {duration: 150}` merges partial options with defaults.
- `transition: null` disables sortable transitions.

## Suggested regression coverage

Add adapter-level tests for each framework binding, or shared tests around the transition resolver if one is introduced:

- omitted `transition` resolves to `defaultSortableTransition`;
- partial transition objects merge with defaults;
- `transition: null` remains `null`;
- updating from an object/default transition to `null` updates `sortable.transition` to `null`.

## Documentation follow-up

React and shared Sortable docs already document `null`. The Solid, Vue, and Svelte sortable pages should also show `transition` as nullable so all framework docs match the core API and public package types.

## Related issues found:

- `#1887` is about `DragOverlay` drop animation transitions, not disabling sortable transitions with `null`.
- `#2038` reports transitions only happening once in a Svelte sortable example. That is different from this cross-framework nullable sortable transition API issue.

No direct duplicate was found in open or closed issues.

NektarTheo:
The React, Vue, Solid and Svelte bindings each merged `defaultSortableTransition` before constructing or updating the core `Sortable`, so an explicit `transition: null` was replaced by the default transition object and items kept animating. `null` is the documented and typed sentinel for disabling sortable transitions, and the core already supports it: the constructor default only applies to `undefined`, and `Sortable.animate()` returns early on a falsy transition.

Added a `resolveSortableTransition` helper to `@dnd-kit/dom/sortable` and used it in all four bindings, so that:

- an omitted transition resolves to `defaultSortableTransition`
- a partial transition merges with the defaults
- an explicit `null` is preserved

The `Disable transition` story previously faked this with `{duration: 0}`; it now uses `transition: null`. The story id is unchanged.

Fixes #2077

### DIFF
--- a/.changeset/tidy-pugs-obey.md
+++ b/.changeset/tidy-pugs-obey.md
@@ -1,0 +1,9 @@
+---
+'@dnd-kit/dom': patch
+'@dnd-kit/react': patch
+'@dnd-kit/vue': patch
+'@dnd-kit/solid': patch
+'@dnd-kit/svelte': patch
+---
+
+Respect `transition: null` in the React, Vue, Solid and Svelte sortable bindings. The bindings previously merged `defaultSortableTransition` before constructing or updating the core `Sortable`, which replaced an explicit `null` with the default transition, so sortable items kept animating. They now use the new `resolveSortableTransition` helper exported from `@dnd-kit/dom/sortable`, which preserves `null` while still merging partial transitions with the defaults.
--- a/apps/docs/docs/solid/hooks/use-sortable.mdx
+++ b/apps/docs/docs/solid/hooks/use-sortable.mdx
@@ -118,8 +118,9 @@ export default function App() {
   An array of plugin descriptors for per-entity plugin configuration. Use `Plugin.configure()` to create descriptors. For example, `Feedback.configure({ feedback: 'clone' })`.
 </ParamField>
 
-<ParamField path="transition" type="SortableTransition" optional>
-  Animation transition configuration.
+<ParamField path="transition" type="SortableTransition | null" optional>
+  Animation transition configuration. Set to `null` to disable sortable
+  transitions entirely.
 </ParamField>
 
 <ParamField path="modifiers" type="Modifier[]" optional>
--- a/apps/docs/docs/svelte/primitives/create-sortable.mdx
+++ b/apps/docs/docs/svelte/primitives/create-sortable.mdx
@@ -129,8 +129,9 @@ Input properties are plain values that are read reactively. To keep an option in
   An array of plugin descriptors for per-entity plugin configuration. Use `Plugin.configure()` to create descriptors. For example, `Feedback.configure({ feedback: 'clone' })`.
 </ParamField>
 
-<ParamField path="transition" type="SortableTransition" optional>
-  Animation transition configuration for sort operations.
+<ParamField path="transition" type="SortableTransition | null" optional>
+  Animation transition configuration for sort operations. Set to `null` to
+  disable sortable transitions entirely.
 </ParamField>
 
 <ParamField path="modifiers" type="Modifier[]" optional>
--- a/apps/docs/docs/vue/composables/use-sortable.mdx
+++ b/apps/docs/docs/vue/composables/use-sortable.mdx
@@ -119,8 +119,9 @@ All input properties accept plain values or Vue refs/getters (`MaybeRefOrGetter`
   An array of plugin descriptors for per-entity plugin configuration. Use `Plugin.configure()` to create descriptors. For example, `Feedback.configure({ feedback: 'clone' })`.
 </ParamField>
 
-<ParamField path="transition" type="MaybeRefOrGetter<SortableTransition>" optional>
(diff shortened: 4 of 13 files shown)


clauderic (approved):
Thanks!

